PR #42, opened 2026-07-02
add option ssh_key, removed power_state

bablokb:
## What this changes

- Improved templating (added optional `ssh_key`)
- removed section `power_state`: the system automatically reboots after i2c and spi are set via rpi.interfaces

This need not be the final version, since some configs (e.g. pairing_code, device_token) could also be sourced from meta-data. Also, there is room for improvement for the mqtt section, e.g. filling the instance-id into client_id and/or device_id.

But for now, this should be "good enough".


dmellok:
Thank you for contributing!

PR #50, opened 2026-07-04
add tesserae to 'i2c' group as well

bablokb:
## What this changes

As the title says. I had to add the tesserae user to `i2c`, otherwise reading of the EEPROM on the Inky-4 failed.

### DIFF
--- a/scripts/pi-client-cloud-init.yaml
+++ b/scripts/pi-client-cloud-init.yaml
@@ -70,6 +70,7 @@ users:
     groups:
       - sudo
       - gpio
+      - i2c
       - spi
     sudo:
       - "ALL=(ALL) NOPASSWD:ALL"


bablokb:
I did not open appropriate issues in the client repos. The manual installs in pi-png and pi-bin probably also need this fix.

dmellok:
Fixed in both client repos: [tesserae-device-pi-png](https://github.com/dmellok/tesserae-device-pi-png) and [tesserae-device-pi-bin](https://github.com/dmellok/tesserae-device-pi-bin). Added `i2c` to the `usermod -aG` step in `scripts/install.sh` and updated the manual-install README sections to match. Same shape as your fix here, so anyone taking the manual path gets the group set that `inky.auto()` actually needs for the EEPROM read.

Issue #52, opened 2026-07-04
[Bug]: Various UI issues


bablokb:
### Tesserae version

0.69.1

### How is Tesserae running?

Other (describe in steps)

### Operating system

Debian Trixie on MicroCloud Container

### Panel hardware

n.a.

### Client repo + version

n.a.

### Steps to reproduce

Various findings and suggestions, mostly cosmetic:

- If a dashboard is linked to multiple devices, only one shows up in the dashboard page
- Timestamps in the history are in UTC, not local time
- The device settings tabs have no save button. There is a single "apply" button for the profile, but that's it. Settings are not saved
- The section "Pi BIN client — tone & dither" shows up two times in the device-settings. Maybe this is not related to this client, but happens because it is the second (of two) devices. This does not happen on my first device of type circuitpython generic
- The weather widget does not use the latitude/longitude from the settings.
- If I change the weather widget type, I have to go through the location search again. Not a big deal, it is fast, but unnecessary
- During navigation in the UI, I regularly get a popup asking me if I really want to leave the page. Maybe this is a Firefox issue, I do experience this with other UIs too (e.g. the MicroCloud web-interface).
- Suggestion: move Devices from Settings to top-level
- Suggestion move device registration to a distinct top-level section (e.g. Admin-Tasks). Currently, the two parts for discover and manual registration fill the whole screen for "Settings->Devices" and I have to scroll down to see the already active clients.
- Suggestion: the landing page should not be "Send", but a sort of admin-dashboard linking to open tasks (e.g. registration) or to devices that sent a log with level "error" or pushes that failed due to rendering problems.

Otherwise: I tested the UI on a small 8" tablet and I found no problems. Very neat responsive design. Still have to do a test on a smaller smartphone but I don't expect any problems there too.

### Server logs

```shell

```

### Before submitting

- [x] I searched [existing issues](https://github.com/dmellok/tesserae/issues?q=is%3Aissue) and didn't find a duplicate.
- [ ] If I could, I tried reproducing on the latest minor release.

dmellok:
Thanks a lot for the thorough writeup, having repro steps for each item makes a real difference on my end. I'll work through the list tomorrow and come back to you once everything's closed out, or sooner if I hit a question on any of them.

dmellok:
[...]

PR #54, opened 2026-07-04
Fix calendar timezone and all-day date handling

charmmmz:
## Summary
- rebase onto current main and reuse the shared `app.tz_resolve.app_timezone()` helper for calendar server-side date calculations
- make calendar day/week/month compute visible dates from settings.app.timezone instead of UTC
- convert local day/week/month boundaries back to UTC before loading iCal events
- bucket timed events by the configured local date so events near midnight appear on the expected day
- filter calendar_day all-day events using iCalendar exclusive DTEND semantics so a previous-day all-day event does not remain visible the next day
- expand calendar_month all-day events across each covered visible date, again treating date-only DTEND as exclusive

## Not addressed
- calendar_week all-day rendering is intentionally left unchanged in this PR. The week client currently skips all_day events in the timetable body; it is unclear whether that is intentional or an omission, and a proper display likely needs a separate all-day row/strip design for the maintainer to weigh in on.

## Tests
- .venv/bin/python -m pytest tests/test_calendar_widget_timezone.py tests/test_renderer_timezone.py tests/test_app_timezone.py -q
- .venv/bin/python -m ruff check app/calendar_time.py plugins/calendar_day/server.py plugins/calendar_week/server.py plugins/calendar_month/server.py tests/test_calendar_widget_timezone.py tests/test_app_timezone.py
- .venv/bin/python -m ruff format --check app/calendar_time.py plugins/calendar_day/server.py plugins/calendar_week/server.py plugins/calendar_month/server.py tests/test_calendar_widget_timezone.py tests/test_app_timezone.py

### DIFF
--- a/app/calendar_time.py
+++ b/app/calendar_time.py
@@ -4,40 +4,6 @@ from __future__ import annotations
 
 from datetime import UTC, date, datetime, time, timedelta, tzinfo
 from typing import Any
-from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
-
-from flask import current_app
-
-from app.tz_resolve import _resolve_iana_timezone
-
-
-def calendar_timezone() -> tzinfo:
-    """Return the timezone calendar widgets should use for date boundaries.
-
-    Rendering already forwards ``settings.app.timezone`` to Chromium so
-    client-side ``new Date()`` uses the configured zone. Calendar server-side
-    fetchers also need the same zone for "today", visible week/month windows,
-    and event date buckets.
-    """
-    raw = "system"
-    try:
-        store = current_app.config.get("SETTINGS_STORE")
-        if store is not None:
-            raw = str((store.get_section("app") or {}).get("timezone") or "system").strip()
-    except Exception:
-        raw = "system"
-
-    if raw and raw.lower() != "system":
-        try:
-            return ZoneInfo(raw)
-        except ZoneInfoNotFoundError:
-            pass
-
-    resolved = _resolve_iana_timezone("system")
-    if resolved:
-        return ZoneInfo(resolved)
-
-    return datetime.now().astimezone().tzinfo or UTC
 
 
 def local_midnight_utc(day: date, zone: tzinfo) -> datetime:
--- a/app/calendar_time.py
+++ b/app/calendar_time.py
@@ -1,0 +1,61 @@
+"""Timezone helpers shared by the calendar widgets."""
+
+from __future__ import annotations
+
+from datetime import UTC, date, datetime, time, tzinfo
+from typing import Any
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+
+from flask import current_app
+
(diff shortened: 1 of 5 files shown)


dmellok:
Thanks for jumping on this. Diagnosis matches what I'd found on a quick pass, and the fix pattern lines up nicely with how `app/scheduler.py` and `app/push.py` handle TZ elsewhere. Tests pass locally for me (13 passed), and the Asia/Hong_Kong + 18:30 UTC freeze case is exactly the boundary that exposes the bug.

You've actually fixed a second bug while you were in there: the all-day DTEND bucketing. The old `s.split("T")[0]` only put each event under its start date, so multi-day all-day events were quietly disappearing after day 1. Good catch even if it wasn't the target.

A couple of small requests if you're up for them:

1. **Would you mind renaming the helper?** `calendar_timezone()` reads calendar-specific, but the logic (read `settings.app.timezone`, fall back through `_resolve_iana_timezone`) is fully general. Something like `app/tz.py` with a `current_timezone()` or `app_timezone()` would let me route another timezone bug through it too, there's a separate report at #52 item 2 (history timestamps in `app/history_routes.py:143`) with the same class of bug.

2. **Any appetite for sweeping `app/history_routes.py:143` in this PR too?** It uses `datetime.fromtimestamp(ev.timestamp).strftime(...)` which follows the container's TZ, so once the helper is generalisable it's basically a one-line change. Would close two reports in one PR if you're up for it. Totally fine if you'd rather keep this scoped to calendars, I'll take it as a follow-up.

Leaving calendar_week's all-day rendering deferral as-is makes sense, that's a design call not a bug. Happy to open a separate issue for the all-day-strip UI if it's useful.

dmellok:
Quick update: v0.69.6 shipped tonight and includes a broader timezone bug sweep (history rows were UTC on Docker containers, that's now fixed via a new `app_timezone()` helper in `app/tz_resolve.py`). Your calendar-widget fix is still very much wanted, it's the only fix for the specific bug you spotted (server-side calendar boundaries in UTC, ignoring `settings.app.timezone`), and the all-day DTEND bucketing fix is a genuine second bug too.

Small ask if you're up for it: since `app_timezone()` now lives in `app/tz_resolve.py` and does the same thing your `calendar_timezone()` in `app/calendar_time.py` does, could you rebase on main and swap the calendar plugins to `from app.tz_resolve import app_timezone`? Everything else in `app/calendar_time.py` (`local_midnight_utc`, `event_local_date_key`, the all-day expansion helpers) is calendar-specific and worth keeping in the module.

No worries if you'd rather just rebase and leave the dedupe for me to sweep after merge, either way's fine.

dmellok:
Thanks for the rebase and the swap to `app_timezone` :) Verified locally, ruff / mypy / pytest all green on `b7f6303`, no conflicts with main. Happy to merge whenever you consider it ready, or say if there's anything else you want to land in this PR first.

charmmmz:
Ready to merge from my side. Thanks!

PR #88, opened 2026-07-09
Fix Weather Now current UV metric

### LINKED ISSUE #85: Weather Now shows daily max UV/rain as current-looking metrics
### What I noticed

On the bundled `Weather, Now` widget, the small metric row currently presents `Rain` and `UV Index` alongside current humidity and wind. At a glance they read like current conditions, but the values appear to come from daily maximum forecast fields.

For example, at night in Shenzhen the dashboard showed UV around `9.1`, which looked suspicious because it was already after dark locally.

A direct Open-Meteo check for the same coordinates returned:

```json
{
  "timezone": "Asia/Shanghai",
  "current": {
    "time": "2026-07-09T03:15",
    "is_day": 0,
    "uv_index": 0.0
  },
  "daily": {
    "uv_index_max": [9.05]
  }
}
```

So timezone handling looks correct. The widget is using the daily max UV value, not the current UV value.

### Code path I found

In `plugins/weather_now/server.py`, the Open-Meteo request asks for:

```text
daily=temperature_2m_max,temperature_2m_min,uv_index_max,sunrise,sunset,precipitation_probability_max
```

and then the metric grid uses:

```python
_metric("Rain", rain_chance, "%", "rainprob", "blue")
_metric("UV Index", _first(daily.get("uv_index_max")), "", "uv", "yellow")
```

where `rain_chance` is `daily.precipitation_probability_max`.

### Question

Is this intentional, as in these two tiles are meant to show today's peak rain/UV risk, while humidity and wind are current conditions?

If yes, maybe the labels should say something like `Rain max` / `UV max` or `Today UV` so it is clear, especially at night.

If no, I think `UV Index` should probably use Open-Meteo's `current=uv_index`, and rain may need a current/current-hour precipitation probability field rather than `daily.precipitation_probability_max`.

Happy to help with a small PR either way if you want: either relabel the daily-risk metrics, or switch the widget to current/current-hour values.

charmmmz:
## Summary

- Request `uv_index` in the Open-Meteo `current=` block for `weather_now`.
- Display `UV Index` from `current.uv_index` instead of `daily.uv_index_max`, so night-time UV correctly renders as 0.
- Keep rain probability as the daily max forecast, but relabel it to `Rain (today)` so it no longer reads like a current condition.

Fixes #85.

## Root cause

The Weather Now metric row mixed current conditions (`Humidity`, `Wind`) with daily maximum forecast fields (`Rain`, `UV Index`) under labels that looked current. `UV Index` was especially visible at night because Open-Meteo returned `current.is_day = 0` and `current.uv_index = 0.0`, while the widget displayed `daily.uv_index_max`.

## Tests

- `./.venv/bin/python -m pytest plugins/weather_now/tests/test_smoke.py -q`
- `./.venv/bin/python -m pytest -q`
- `./.venv/bin/ruff check .`
- `./.venv/bin/ruff format --check .`
- `./.venv/bin/python -m mypy app` did not reach project code locally: mypy exited while parsing the installed numpy stub because the repo config targets Python 3.11 and the local numpy stub uses Python 3.12 `type` statement syntax.

### DIFF
--- a/plugins/weather_now/server.py
+++ b/plugins/weather_now/server.py
@@ -86,7 +86,8 @@ def fetch(
         f"?latitude={lat}&longitude={lon}"
         "&current=temperature_2m,weather_code,apparent_temperature,"
         "wind_speed_10m,wind_direction_10m,relative_humidity_2m,is_day,"
-        "dew_point_2m,wind_gusts_10m,cloud_cover,pressure_msl,visibility"
+        "dew_point_2m,wind_gusts_10m,cloud_cover,pressure_msl,visibility,"
+        "uv_index"
         "&daily=temperature_2m_max,temperature_2m_min,"
         "uv_index_max,sunrise,sunset,precipitation_probability_max"
         f"&temperature_unit={temp_unit}"
@@ -122,6 +123,7 @@ def fetch(
 
     speed_unit = "mph" if units == "imperial" else "km/h"
     rain_chance = _first(daily.get("precipitation_probability_max"))
+    uv_index = current.get("uv_index")
 
     # The 8-metric grid the Refined / Geometric / Swiss variants paint.
     # Order is hand-tuned: most-useful first so a 4-cell variant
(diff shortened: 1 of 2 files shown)


dmellok:
LGTM! Thank you.

PR #93, opened 2026-07-10
Surface environmental telemetry in device status

charmmmz:
## Summary

- promote `temperature_c` and `humidity_pct` to well-known ESP32 heartbeat fields with float coercion
- preserve both values across partial/offline heartbeats using the existing status merge behavior
- treat environmental drift as volatile so temperature and humidity changes alone do not create event-log noise
- show available readings in an optional Environment tile in the existing device Status area
- keep Celsius as the canonical telemetry unit; widgets can convert units at presentation time

This implements the server half of the approach agreed in [Discussion #89](https://github.com/dmellok/tesserae/discussions/89). The matching firmware change is https://github.com/dmellok/tesserae-device-firmware/pull/1.

## Validation

- `pytest -q`: 1516 passed
- `ruff check .`: passed
- `ruff format --check .`: passed
- GitHub Actions `ruff`, `mypy (strict modules)`, and `pytest`: passed
- desktop and 390px mobile browser checks passed with no tile overflow or overlap
- temporarily hot-patched a Tesserae 0.74.0 container and received a real E1004 heartbeat containing `temperature_c: 32.1031` and `humidity_pct: 77.7625`
- the device remained registered and continued its normal frame/status/300-second sleep cycle

### DIFF
--- a/app/settings/index_routes.py
+++ b/app/settings/index_routes.py
@@ -10,6 +10,7 @@ shape regardless of source. Helpers below (``_values_for_core``,
 from __future__ import annotations
 
 import json
+import math
 import os
 import time
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


dmellok:
Reviewed and validated, all green. Thank you for the contribution — merging.

PR #96, opened 2026-07-10
feat(status): show device temperature and humidity in the status bar

charmmmz:
## Summary

- show a panel's latest temperature and humidity in the built-in `tesserae_status` widget
- render the readings as one compact per-device environment cluster with metric / imperial presentation
- keep the 48 px bar usable on narrower panels and fix the editor's duplicate-looking status-bar toggle

## Context

This is the UI follow-up discussed in [Discussion #89](https://github.com/dmellok/tesserae/discussions/89).

The data path is already available through the merged firmware and server work:

- [tesserae-device-firmware#1](https://github.com/dmellok/tesserae-device-firmware/pull/1) publishes `temperature_c` and `humidity_pct`
- [#93](https://github.com/dmellok/tesserae/pull/93) promotes those fields in `parse_status()`, treats them as volatile telemetry, and surfaces them in device status

This PR consumes those canonical fields without changing the wake order or triggering any additional render.

## Behavior

- Per-device pushes read temperature and humidity from the render's `target_device_id`, matching the existing battery and Wi-Fi behavior.
- Editor previews without a target use the most recent single sensor-bearing heartbeat. Values are never averaged or combined across devices.
- Missing or invalid readings auto-hide; humidity is clamped to 0-100%.
- `temperature_c` remains canonical. The widget converts to Fahrenheit only at presentation time through its `metric` / `imperial` option.
- Temperature and humidity can be enabled independently and render as one compact cluster.
- Long dashboard names truncate in bar mode, and spacing tightens below 900 px without hiding enabled chips.
- The editor option order is `Temperature -> Temperature units -> Humidity`.
- The status-bar toggle now opts into the shared `field--switch` styles, hiding the native checkbox so only the custom switch track is visible.

## Validation

- `.venv/bin/pytest -q`: **1536 passed**
- focused status/page tests: **65 passed**
- `.venv/bin/ruff check .`
- `.venv/bin/mypy --python-version 3.12 app`: **101 source files**
- `node --input-type=module --check < plugins/tesserae_status/client.js`
- `git diff --check`
- verified with real SHT4x telemetry on a reTerminal E1004
- checked 800 px and 1200 px bar layouts, long dashboard names, Celsius, Fahrenheit, and text-only mode
- hot-patched a 0.74.0 deployment and verified the actual dashboard editor and 1200x1600 compose output


dmellok:
This is a clean one, thanks. It reuses the same `DEVICE_STATUS` heartbeat the battery and wifi chips already read, so no new data path or egress, and the guards are all right: finite checks, humidity clamped, canonical Celsius with °F only at display. Using the single most-recent sensor heartbeat instead of averaging across panels is the right call.

Thanks too for the tests and the real-hardware check on the E1004. Merging.

Cheers,
Kayden

PR #135, opened 2026-07-21
Add fetch_latest button action and History preview

charmmmz:
## Summary

Companion firmware change: [dmellok/tesserae-device-firmware#5](https://github.com/dmellok/tesserae-device-firmware/pull/5).

- add a `fetch_latest` button action that returns the latest existing frame without rendering, pushing, changing rotation state, or setting a manual override
- accept the deployed firmware's legacy `event` frame-query parameter as a fallback alias for `button_event_id`
- snapshot the current `composition_digest` on fetched History rows so they have a thumbnail and click-through
- separate History previewability from resendability, keeping Resend disabled when the event's top-level `digest` is null
- document the action and client protocol, including the preferred query parameter and compatibility alias

## Why

There was no button action for re-downloading a frame the server had already rendered: `refresh` necessarily enters the render/push pipeline.

In addition, firmware through v1.5.0 sends `event` on `/frame`, while the server only parsed `button_event_id`. The event id was therefore dropped on that path and deduplication depended on the short time window.

Finally, button History rows deliberately used `digest=null` to disable Resend, which also prevented a fetched row from showing the existing composition it downloaded. This keeps `digest` as the resend contract and adds a preview-only `composition_digest`.

## Confirmed contract

This implements the server half agreed in [Discussion #132](https://github.com/dmellok/tesserae/discussions/132#discussioncomment-17714414):

- `button_event_id` is preferred and `event` remains a fallback alias
- `fetch_latest` bypasses a matching ETag for that response without producing a new render
- fetched rows carry the existing composition preview while Resend stays disabled

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/pytest -q` — 2118 passed
- `git diff upstream/main...HEAD --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,11 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Added
 
+- **`fetch_latest` button action.** Re-downloads and repaints the latest frame
+  already rendered for the device without running the composer, publishing a
+  new artefact, moving the rotation, or setting a manual override. The action
+  bypasses a matching `If-None-Match` for its `/frame` response, so its meaning
+  does not depend on a particular firmware clearing its cached ETag first.
 - **MCP: rotations, schedules, and decks.** The agent MCP surface (`/api/mcp`) and the
   `tesserae-mcp` bridge now expose list / create / delete for rotations and schedules, and list /
   create / delete / suggest for decks, so an agent that builds pages can also wire how they cycle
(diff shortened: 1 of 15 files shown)


dmellok:
Reviewed. Clean and well-scoped — verified the three pieces hold together: `fetch_latest` bypasses the matching ETag without touching the render pipeline (`_latest_renders` left intact), the `event` alias only falls back when `button_event_id` is absent so the canonical id always wins, and the preview/resend split keeps `digest` null while surfacing the composition thumbnail. Docs (client-protocol, openapi, install) and tests all updated. One note for later: `event` is now a reserved query name on `/frame`. LGTM.

PR #139, opened 2026-07-24
fix: distinguish firmware queue from rollout membership

charmmmz:
## Summary

- distinguish retained canary/fleet rollout membership from an update that is still pending
- show the imported release version in the queued chip instead of a potentially newer online `Available` version
- keep completed canaries available to the Fleet rollout model while rendering their device row as up to date
- add regression coverage for pending, completed, promoted, and newer-online-release states
- document the fix and bump the app version to 0.188.1

## Root cause

The unified Firmware page treated `is_canary` as equivalent to `queued`. Because rollout membership is intentionally retained, a device that had already reported the imported release still rendered as queued. The delivery path itself was safe: `_released_ota()` already withholds releases that are not strictly newer than the device's reported firmware.

The device-list model now derives effective queue state from both membership and whether the imported release is still newer. It keeps the underlying membership intact for the Advanced/Fleet view.

This also prevents an older imported release membership from making a newer online `Available` version look queued when that newer descriptor has not actually been imported.

## Validation

- `python -m pytest tests/test_firmware_rollout.py -q` — 32 passed
- `python -m pytest -q` — 2257 passed
- `python -m ruff check .` — passed
- `python -m ruff format --check .` — passed

Local `mypy app` was blocked before project analysis by the shared Python 3.12 environment's NumPy 2.5.1 stubs being parsed against this project's Python 3.11 mypy target. The canonical clean Python 3.11 mypy job remains enabled on this PR.

## Context

Real-device reproduction and the proposed state boundary were discussed here:

https://github.com/dmellok/tesserae/discussions/121#discussioncomment-17754716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Completed firmware updates stayed labelled as queued.** The Firmware page now distinguishes
+  retained canary/fleet rollout membership from a release that is still newer than the device.
+  Once a device reports the imported version it reads as up to date, while pending rows name the
+  actual imported release instead of a potentially newer online Available version.
+
 - **Deck heartbeat reports could revert a freshly pushed dashboard.** v0.187.1's report promotion
   promoted whatever page the panel said was on glass, so a heartbeat arriving between a push and
   the device's next fetch clobbered the pending frame with the older deck frame, and the new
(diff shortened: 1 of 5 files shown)


dmellok:
Merged as 9159a52, shipping in v0.189.1 (renumbered from 0.188.1 during conflict resolution; main had moved to 0.189.0 under you, which was also why the branch showed conflicts, both were mechanical). The fix itself was exactly right: deriving the queued chip from membership plus the same is-newer gate `_released_ota()` applies means the UI can't disagree with what delivery would actually offer, and keeping membership intact for the Fleet view was the correct split. The four regression tests slotted straight into the suite (2261 green on merge). Thanks for the clean root-cause writeup and for linking the real-device repro from the #121 thread; it made review fast.

PR #148, opened 2026-07-28
Fix photo orientation for portrait-native ESP32 panels

charmmmz:
## What changed

- Fit arbitrary Send-page and Companion image uploads into the display's composition dimensions before reconciling that composition with the firmware-native row stride.
- Base the ESP32 packer's 90-degree rotation on `panel.w × panel.h` versus `native_w × native_h`, rather than on the uploaded image's own aspect ratio.
- Add Fit and Fill regressions for a 4:3 landscape photo sent to a portrait-native panel.
- Document the user-visible fix in the changelog.

## Why

This surfaced while testing the community iOS Companion with a portrait Seeed reTerminal E1004 (1200 × 1600). A 4:3 landscape photo appeared upright in Tesserae History, but the physical display painted it 90 degrees clockwise. The same bug also affected direct uploads through the web Send page because both paths use `PushManager.push_image()` and the same `esp32_bin` transform.

History was correct because it preserves the source composition before the device-specific transform. In `esp32_bin`, the renderer compared the raw upload's orientation with the firmware-native dimensions and rotated whenever they differed. That treated a landscape photo as if it were a landscape-calibrated panel composition.

The corrected pipeline first applies the requested Fit/Fill mode at the display's composition dimensions. It then rotates only when the completed composition orientation differs from the firmware-native stride. Existing calibrated portrait/landscape mounting behavior remains covered by the renderer tests.

## User impact

Landscape photos sent to portrait-native ESP32 panels such as the reTerminal E1004 now retain their visual orientation. The fix applies to both the web Send page and Companion clients.

## Validation

- `13 passed` — `renderers/esp32_bin/tests/test_smoke.py`
- `73 passed` — Companion jobs, web Send routes, PushManager, and push extras
- Ruff lint and format checks passed
- Hot-patched and verified on a running Tesserae v0.207.0 instance
- Retested with the original photo on a physical reTerminal E1004; the reporter confirmed the orientation is now correct

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Photo sends to portrait-native ESP32 panels keep their visual orientation.**
+  Send-page and Companion uploads are now fitted into the display's composition
+  dimensions before the renderer reconciles that composition with the
+  firmware-native row stride. Landscape photos sent to devices such as the
+  reTerminal E1004 no longer rotate 90 degrees clockwise while History remains
+  upright.
+
 - **v2 staleness is anchored to layout, not pixels (live bench round 3, 2026-07-25).** A region
   report against a superseded frame digest now dispatches when that frame's untrimmed region-id
   set matches the live one, resolved through a new ~10-generation digest lineage per device; a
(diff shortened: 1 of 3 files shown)


dmellok (approved):
Traced the rotation path and this is right. Keying the 90° off the raw upload's aspect was the bug; fitting into the composition dims first and rotating only on a composition-vs-native-stride mismatch is the correct invariant, and it leaves page pushes byte-identical because the fit no-ops when the frame already arrives at panel dims. Both calibrated rotation directions stay covered by the retained `*_calibration_on_*` tests, and the two new fit/fill regressions lock the photo path. CI green, real-device confirmed on the E1004. Merging. Thanks for the clean writeup and the physical retest.

PR #149, opened 2026-07-28
Fix battery analytics during active charging

charmmmz:
## Summary

- separate an active charging phase from the latest clean discharge phase
- expose live `is_charging` / `charge_rate_per_day` values while retaining a historical `slope_per_day` only for drain data
- replace OLS fits with a Theil-Sen median slope, deterministically subsampling long windows to 300 points
- update the Device batteries page and JSON endpoint to show `Charging`, `Charge rate`, `Full in`, and `Last drain rate`
- keep every raw sample unchanged in the chart

## Root cause

`_last_discharge_segment()` treated every sample after a charge jump as discharge. With frequent telemetry during an active charge, that suffix is the rising charge ramp, so the UI could render a large positive value under `Drain rate`.

The inverse `_last_charging_segment()` only stopped on drops greater than one percentage point. At a five-minute cadence, it could absorb a long preceding discharge or plateau and dilute the time-to-full estimate.

## Behavior

While charging:

- the trailing rising phase supplies `Charge rate` and `Full in`
- the preceding clean discharge phase supplies `Last drain rate`, when available
- discharge projections stay blank
- a positive slope can never render under a drain label

While flat or discharging, the existing drain labels and projections remain in place.

## Validation

- `2504 passed` in the full test suite
- `33 passed, 75 deselected` in the focused battery/widget/REST selection
- Ruff formatting and lint checks
- mypy on the changed application modules
- replayed the original 1,957-sample E1004 history through the patched container: the 82% charge ramp classified as charging, with separate live charge and historical drain rates

## Discussion

Implements the semantics and test coverage agreed in:

- https://github.com/dmellok/tesserae/discussions/137
- https://github.com/dmellok/tesserae/discussions/137#discussioncomment-17742398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Battery analytics now separate charging from drain phases.** The Device batteries page
+  shows live `Charging` / `Charge rate` / `Full in` values without labelling a positive charge
+  ramp as `Drain rate`, and retains the preceding clean discharge slope as `Last drain rate`
+  when available. Robust Theil-Sen fits prevent a single glitched reading or post-unplug voltage
+  relaxation drop from dominating either estimate, while raw chart samples remain unchanged.
+
 - **v2 staleness is anchored to layout, not pixels (live bench round 3, 2026-07-25).** A region
   report against a superseded frame digest now dispatches when that frame's untrimmed region-id
   set matches the live one, resolved through a new ~10-generation digest lineage per device; a
--- a/app/device_battery_routes.py
+++ b/app/device_battery_routes.py
@@ -143,6 +143,8 @@ def _device_card(
         "series": series,
         "prediction": {
             "slope_per_day": prediction.slope_per_day,
+            "is_charging": prediction.is_charging,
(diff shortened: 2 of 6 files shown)


dmellok (approved):
Reviewed the algorithm end to end and it matches the semantics we agreed in #137. The phase split is right: the trailing charge segment drives `is_charging` / `charge_rate_per_day` / `Full in`, the preceding clean discharge supplies `Last drain rate`, and a positive ramp can never surface under a drain label (`slope_per_day` is clamped to `min(0, slope)` in the flat case and taken only from the prior discharge while charging). Theil-Sen with the deterministic 300-point subsample is the right robustness call, and starting the charge segment at the high side of a jump keeps an instantaneous firmware correction from inflating the rate.

Checked the one thing that could have bitten: `slope_per_day` becoming `float | None` is threaded through cleanly, both the `Drain rate` and new `Last drain rate` tiles guard `is not none`, so a charge with no prior clean discharge renders rather than 500s. The three regression scenarios are exactly the ones I asked for. 2504 green, real E1004 history replayed.

One forward-compat note for later: `/device-batteries/<id>/series.json` can now return `slope_per_day: null`, so any consumer doing math on it has to handle null. Merging.

PR #153, opened 2026-07-29
Add Companion 0.4 History and photo layout modes

charmmmz:
## Summary

Implements the first server batch of the accepted Companion 0.4 contract from
[Discussion #147](https://github.com/dmellok/tesserae/discussions/147):

- advertise and accept all five server image layout modes (`fit`, `fill`,
  `blur`, `stretch`, and `center`);
- expose paginated canonical push History from the existing `EventLog`;
- serve retained composition PNGs with ETag revalidation;
- idempotently resend retained entries to their original target snapshot while
  respecting per-device quiet hours;
- correlate successful dashboard, photo, and resend jobs to exact
  `history_event_ids` when available;
- preserve the original fit mode and target snapshot on photo resend.

The OpenAPI 0.4 contract and fixtures are synced byte-for-byte from the
[community iOS Companion PR](https://github.com/charmmmz/tesserae-companion-ios/pull/1).

## Why

The iOS app currently has a local Activity feed, but it cannot recover server
History after reinstalling or observe pushes made by schedules, the web UI, or
other clients. Using Tesserae's existing `EventLog` as the canonical source
gives all clients one honest timeline without introducing a second store.

Resend reuses the retained composition rather than re-rendering its source. This
keeps the operation faithful, fast, and suitable for an iOS background job, while
the idempotency key prevents an uncertain retry from refreshing an e-ink panel
twice.

## Compatibility and boundaries

- History is additive and advertised through the `history` feature.
- `limits.image_fit_modes` is optional for older clients; clients that do not
  see it can continue to assume Fit and Fill.
- Existing persisted jobs remain readable because `history_event_ids` is
  optional.
- Existing web History resend behavior is unchanged when no target subset is
  supplied.
- This batch intentionally serves the retained composition for History preview.
  A device-final, post-fit preview for the Displays screen remains the agreed
  follow-up batch and is not implemented here.

## Validation

- `2563 passed` — full Tesserae test suite
- Ruff check and format check
- strict mypy on the changed Companion, History, job-store, EventLog, and push
  modules
- `git diff --check`
- vendored OpenAPI and fixtures compared byte-for-byte with the iOS repository

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Added
 
+- **Companion API 0.4 adds canonical History and all five photo layout modes.**
+  Community clients can page through the same push History shown in the web UI,
+  fetch retained composition previews, and idempotently resend an entry to its
+  original display snapshot while respecting quiet hours. Successful jobs may
+  include the exact new History event IDs, and image uploads now advertise and
+  accept Fit, Fill, Blur, Stretch, and Center.
+
 - **The opt-in daily heartbeat now includes a bucketed count of paired companion apps.** Same
   `0`/`1`/`2-3`/… bucketing as the device count, derived server-side from the live companion
   tokens (never a client name, install id, or app version), so adoption of the community
--- a/app/companion_api.py
+++ b/app/companion_api.py
@@ -10,14 +10,14 @@ or Flask session cookies.
 
 This module covers the discovery, pairing, session, read-model, and write
 surfaces of the contract: capability probe, pairing, session revocation,
-device + dashboard read models, the two async write routes (dashboard
-push, image push), and job polling. Writes return ``202`` with a persisted
+device + dashboard read models, canonical History reads and resend, the
+async write routes, and job polling. Writes return ``202`` with a persisted
 job the client polls; quiet-hours suppression surfaces as a *successful*
 ``quiet`` outcome, never a failure. ``_tesserae._tcp`` discovery lives in
 app.mdns.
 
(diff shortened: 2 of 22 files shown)


dmellok (approved):
Reviewed end to end, including the security surface, and it's a clean approve.

The additions are additive and backward-compatible: `republish()` keeps its old behaviour when `device_ids` is omitted, `history_event_ids` is optional so persisted jobs are unchanged, and `before_id` is parameterised in the EventLog query.

Security holds up:

- The preview artifact lookup rejects traversal, malformed digests, and symlink escape (`_DIGEST_RE` + `resolve(strict=True)` + `relative_to(root)`).
- Resend binds strictly to the original target snapshot and re-intersects inside `republish`, so it can never reach a device outside the original push.
- The history cursor is int-validated (`parse_history_id`) and parameterised in SQL.
- Preview-only rows are viewable but never resendable.
- Scopes are enforced on all three routes.

Test coverage for the security cases (traversal, cursor, target-binding, pruning, non-resendable) is thorough, and the vendored 0.4 contract/fixtures match the client repo.

Two small things I'll note rather than block on:

- History reads under `devices:read`. Fine for the single companion role, but worth recording as a scope choice since History spans dashboards + activity; `dashboards:read` (or both) would read more naturally if scopes ever narrow.
- `HistoryItem.error` is passed through verbatim. Push error strings are user-safe today; just something to keep in mind if a future error message embeds an internal path.

Approved and merge-ready from my side. Since this is the first batch and still Draft, mark it ready when you consider it complete (or say if more is landing in this batch), and I will bump the version, relock, and merge. Happy to take the follow-up device-final preview batch the same way.

PR #154, opened 2026-07-29
Add logical Companion device previews

charmmmz:
## Summary

- retain a content-addressed logical-screen PNG for each device's latest full render
- apply the selected Fit, Fill, Blur, Stretch, or Center layout in logical panel coordinates
- preserve device underscan and renderer letterbox background without exposing firmware-native rotation, mount flip, or vertical scan compensation
- serve that preview from the existing Companion endpoint with its own ETag and safe retained-file lookup
- backfill previews for pre-upgrade latest-render records on the next unchanged render without repainting the e-ink panel
- keep live, previous, and warmed Deck previews protected by the existing artifact GC

## Why

`GET /api/app/v1/devices/{device_id}/preview` currently serves the original composition PNG. For photo sends that means the iOS Displays card always looks like the source image, even when the display actually used Fill, Blur, Stretch, or Center. It can also misrepresent portrait displays because renderer wire artifacts include firmware-native row-stride and mount compensation that should not appear in a phone preview.

This implements the OpenAPI 0.4 logical device-preview contract discussed in #147. It deliberately stops at accurate fit, geometry, underscan, low-battery overlay, and supported letterbox background. Renderer palette, quantisation, tone, dither, and client-side saturation remain a renderer-specific follow-up rather than being presented as byte-exact on-glass output.

## Dependency

#153 has merged. This branch is rebased onto `main` after its `v0.211.0` version bump, so the PR now contains only the logical device-preview commit.

## Validation

- `pytest -q`: 2582 passed before the dependency rebase
- focused preview/push tests after rebasing onto current `main`: 70 passed
- `ruff check .`
- `ruff format --check .`
- `mypy --python-version 3.12 app/device_preview.py app/push.py app/companion_api.py`
- `git range-diff` confirmed the rebased commit is patch-equivalent to the reviewed implementation
- three independent read-only reviews of the renderer pipeline, OpenAPI/iOS contract, and security/lifecycle behavior found no blocking issues

Local mypy with the repository's configured Python 3.11 target is currently blocked before project code by the installed NumPy stub using Python 3.12 `type` syntax; the changed strict modules pass when checked against the active Python 3.12 environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,14 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Companion display previews now match each display's selected photo layout.**
+  The server retains a separate logical-screen PNG after Fit, Fill, Blur,
+  Stretch, or Center and device underscan are applied, so portrait displays no
+  longer show the unfitted source composition in the iOS app. Hardware-only
+  row-stride rotation and mount compensation remain confined to the device
+  artifact, while preview ETags, upgrade backfill, and artifact pruning track
+  the logical frame independently.
+
 - **The Webpage widget captures JavaScript pages after their data loads.** It was
   screenshotting the moment the embedded page fired its `load` event, before a data-driven
   page (a weather dashboard, an SPA) ran its post-load fetch and painted, so the panel showed a
--- a/app/companion_api.py
+++ b/app/companion_api.py
@@ -51,6 +51,7 @@ from app.companion_history import (
 )
 from app.companion_jobs import CompanionJobs, JobOutcome
 from app.device_loader import Device, DeviceRegistry
+from app.device_preview import retained_device_preview
 from app.panel import device_panel, resolve_settings_panel
 from app.quiet_hours import device_is_quiet
 from app.state.companion_token_store import COMPANION_SCOPES, CompanionTokenStore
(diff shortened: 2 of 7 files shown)


dmellok (approved):
Reviewed end to end, including the security and lifecycle surface, and it's a clean approve.

The logical-preview design is right: fit and underscan run in `panel.w x panel.h`, and `flip` / `vflip` / firmware-native rotation are deliberately omitted, so a photo preview shows the real Fit/Fill/Blur crop upright instead of the packed wire frame, while dashboards stay WYSIWYG (the fit no-ops when the composition is already panel-sized).

Security matches the #153 bar:

- `retained_device_preview` uses the same traversal/symlink-proof lookup (`^[0-9a-f]{16}$` + `resolve(strict=True)` + `relative_to(root)` + symlink reject).
- Fit mode and background are validated against allow-lists, and the preview is content-addressed.
- Same `devices:read` endpoint, now serving `preview_digest`; no new auth surface.

Lifecycle holds up:

- `preview_digest` is added to `_live_digests()`, so previews for live, previous, and warmed-Deck frames survive artifact GC and are collected only when orphaned.
- The preview write is best-effort, so a disk or PNG failure can never turn a successful panel publish into a failed send.
- The low-battery overlay is consistent across both write paths: `_publish_artifact` overlays before the normal-path write, and the unchanged/backfill path re-overlays from the stored composition explicitly.
- The backfill path regenerates a missing preview on an unchanged render without repainting the panel, which is the correct upgrade behaviour.

One non-blocking note: the preview PNG is generated on every render, including on installs with no companion app paired. It is bounded by content-addressed dedupe, the GC, and the best-effort guard, so it is a minor unconditional cost rather than a correctness issue; gating it on companion-in-use is a possible later optimisation.

Approved and merge-ready from my side. It is stacked on #153, so once you mark it ready and rebase onto `main` (so the PR is just the logical-preview commit), I will bump the version, relock, and merge.

PR #158, opened 2026-07-29
fix: preserve Companion History identity

charmmmz:
## Summary

- preserve every canonical History event ID produced by a multi-panel dashboard push
- expose those IDs through the existing terminal Job `history_event_ids` field
- snapshot the originating display ID on button-triggered History rows
- retain the existing singular `event_id` behavior for single-panel callers

## Why

Follow-up to Discussion #147.

The iOS Companion shows an accepted/running Job immediately, then replaces it
with canonical server History once the terminal Job identifies the History rows
it created. `PushManager.push()` renders once per distinct panel and each
`_fan_out()` call records its own canonical push event, but the aggregate
`PushResult` discarded those per-panel event IDs. A successful dashboard Job
therefore omitted `history_event_ids`, leaving clients with both the local Job
and server History card.

Button `fetch_latest` History had a related identity gap: the service knew the
originating device, but did not include it in the row's `device_ids` snapshot,
so Companion clients displayed “No displays”.

This keeps the public API shape unchanged. The new plural collection is
internal to `PushResult`; the Companion Job continues to return the existing
optional `history_event_ids` array.

## Validation

- `pytest tests/test_push.py tests/test_button_service.py tests/companion` — 159 passed
- `ruff check` on all changed Python files
- `ruff format` on all changed Python files
- `mypy --python-version 3.12 app/push.py app/companion_api.py app/button_service.py`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Companion Activity now keeps exact History and display identity.** Multi-panel
+  dashboard pushes preserve every canonical History event ID in their terminal
+  Job, and button-triggered fetch rows snapshot the originating device ID, so
+  clients no longer need title/time heuristics to suppress duplicate Activity
+  cards or identify the display that fetched a frame.
+
 - **Companion display previews now match each display's selected photo layout.**
   The server retains a separate logical-screen PNG after Fit, Fill, Blur,
   Stretch, or Center and device underscan are applied, so portrait displays no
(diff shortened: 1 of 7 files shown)


dmellok:
Reviewed end to end and merged; shipped on edge in v0.213.0. The multi-panel gap was the right root cause: `_push_page_locked` now aggregates every panel group's event id into `PushResult.event_ids` (deduped), and `_history_event_ids` folds that plus the legacy `event_id`, so a multi-panel dashboard push correlates all its canonical History rows to the terminal Job and the client can drop its dedupe heuristics. The button-row `device_ids` snapshot is the right complement, and also makes those rows resendable rather than "No displays". Additive and backward-compatible, no new surface, with tests covering the aggregate, the button snapshot, and event-id resolvability.

PR #162, opened 2026-07-30
feat(companion): track last-served device previews

charmmmz:
## Summary

- record the last frame actually served to each REST device on both `200` and matching `304` fetches
- expose that last-served logical preview through the Companion device preview route
- add a backward-compatible `has_pending_render` field so clients can distinguish the on-glass frame from a newer frame waiting for the next wake
- retain last-served preview artifacts through preview garbage collection and carry the state through render replacement
- update the Companion OpenAPI contract and fixtures to 0.4.1

## Why

A sleeping e-ink device may not have fetched the newest render yet. The existing Companion preview therefore represented the latest server-side frame, not necessarily what was physically on the display. This change gives the app an honest current-screen preview while still indicating when a newer frame is pending.

No firmware change is required: the server already knows which digest it returned from the REST frame endpoint. MQTT/push devices retain the existing latest-render preview behavior because delivery is not acknowledged through this polling path.

This follows the current-screen discussion in #147.

## Contract behavior

- `preview_url` for REST devices resolves to the last-served preview when available
- `has_pending_render` is `true` when the latest queued wire digest differs from the last-served digest
- a matching conditional `304` counts as serving the current frame and clears the pending state
- older state files remain valid; the new persisted fields are optional

## Validation

- `166 passed` across the affected push, REST, Companion API, preview, and contract tests
- Ruff passes for all affected Python files
- `git diff --check` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Added
 
+- **Companion display cards distinguish served and pending frames.** REST
+  `/frame` polls now retain the last frame handed to each device, including
+  matching `304 Not Modified` confirmations. The Companion device preview uses
+  that served frame while `has_pending_render` reports when a newer render is
+  waiting for the panel's next wake; MQTT and push transports continue to use
+  the latest server render.
+
 - **Companion API 0.4 adds canonical History and all five photo layout modes.**
   Community clients can page through the same push History shown in the web UI,
   fetch retained composition previews, and idempotently resend an entry to its
--- a/app/companion_api.py
+++ b/app/companion_api.py
@@ -17,7 +17,7 @@ job the client polls; quiet-hours suppression surfaces as a *successful*
 app.mdns.
 
 Contract: ``charmmmz/tesserae-companion-ios`` ``Contracts/app-v1.openapi.yaml``
-(OpenAPI 0.4.0). The vendored copy + fixtures under ``tests/companion/``
+(OpenAPI 0.4.1). The vendored copy + fixtures under ``tests/companion/``
 guard these shapes.
 
 mypy --strict applies to this module, see pyproject.toml.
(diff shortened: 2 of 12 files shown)


dmellok:
Reviewed and ran the full affected suite locally on this branch: 166 passed across the push, REST, Companion API, preview, and contract tests, plus `mypy --strict` clean and ruff clean.

What I checked beyond the tests:

- **Serve-path coverage.** `get_frame`'s button/touch dispatch falls through to the single 304/200 pair rather than returning early, so `record_frame_served` lands on every frame-serving path, including `?button=refresh` re-renders and force-download. The 204 no-frame path correctly records nothing.
- **No desync path.** Folding the three former inline `_latest_renders[dev] = dict(info)` sites (fan-out, deck promote, shadow promote) into a single `_replace_latest_render_locked` is the right call: after this there are only two writers left, and a promotion changes "what to serve next" without rewriting the last-served snapshot. The invariant holds.
- **The mid-response race** is handled the way the docstring describes: recording the captured digest onto the current entry means a render that lands while the response is assembled correctly reports `has_pending_render` rather than falsely clearing it.
- **GC.** Adding `last_served_preview_digest` to `_live_digests` keeps the served preview PNG through prune; leaving the served `.bin` unprotected is correct since it's never re-served.
- **Contract.** `has_pending_render` optional + `additionalProperties: false` is properly additive, and the spec/fixtures/`_schema.py` bump to 0.4.1 is consistent.

One observation, not a change request: `has_pending_render` / `last_served_render_for` read `_latest_renders` without the lock, so a concurrent in-place `record_frame_served` can yield a momentarily stale pending bool. That matches the existing lock-free `latest_render_for` convention, so it's consistent with the file.

Looks good to me. Holding the merge until you take it out of draft.

dmellok (approved):

dmellok:
Merged via a local merge commit rather than the button, so a quick note on why.

CI here went red on the pytest jobs, but it was not this PR. `test_pre_04_job_without_history_ids_remains_compatible` had a fixture with a hard-coded timestamp that aged past the 24h job-retention window once the date rolled over, so it started failing on `main` independently (fixed in cce8f02). This branch predated that fix, so its CI inherited the failure. Your changes were clean: I ran the affected push/REST/Companion/preview/contract suites plus the full suite locally and they pass.

Merging locally let me pull that test fix in and resolve the version bump in one step (`main` had already advanced past this branch's fork point), so `main` lands green at 0.213.4 with the last-served preview work intact. Nothing needed from you; the contract is now at 0.4.1 on the served copy under `tests/companion/`.

PR #173, opened 2026-07-31
feat(companion): expose dashboard Phosphor icons

charmmmz:
## Summary

- include each dashboard’s optional bare Phosphor `icon` identifier in `GET /api/app/v1/dashboards`
- update the mirrored Companion contract to OpenAPI 0.5.1
- cover both a normalized icon value and the nullable fallback in fixtures/tests

## Why

The iOS Companion can now use the same dashboard icon identifiers as the web UI instead of maintaining a separate SF Symbols mapping. Older clients remain compatible because the field is additive and optional.

## Testing

- `python -m pytest tests/companion` — 122 passed
- `python -m ruff check app/companion_api.py tests/companion/test_companion_api.py tests/companion/test_contract_fixtures.py`
- mirrored iOS OpenAPI contract and fixture are byte-identical to this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Added
 
+- **Companion dashboard listings now include their Phosphor icon name.** The
+  optional `icon` field uses the same bare identifier as the web dashboard
+  list, so companion clients can render a consistent icon and safely fall back
+  when a dashboard has no icon.
+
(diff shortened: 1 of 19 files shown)


dmellok:
Merged; this lands on EDGE as v0.217.0, and the vendored contract copy on main is now 0.5.1.

One related note from today's work on main: dashboard `icon` values are bare Phosphor slugs from the same vendored manifest the MCP `/icons` search uses, so the identifiers this field returns should always resolve on your side. The server also now flags unresolvable icon references in `render_report` (`icon_invalid`), which may be useful while wiring the iOS icon mapping.

PR #182, opened 2026-08-03
Generalize Apple Reminders personal-data contract

charmmmz:
## Summary

- add the strict grouped `reminders` personal-data source for up to 20 selected lists and 200 incomplete items in aggregate
- advertise accepted personal-data source IDs through optional `personal_data.sources` instead of adding a feature flag per schema
- keep the `personal_data_bridge_v1` envelope and the existing latest-only, required-expiry semantics
- retain `reminders.fridge` and `personal_data_reminders` only for the already published Reminders, Fridge widget
- add synthetic request/response fixtures and contract guards

## Why

This follows the accepted direction in [Discussion #176](https://github.com/dmellok/tesserae/discussions/176#discussioncomment-17874174): Apple Reminders is the source domain, while fridge, tasks, and shopping are widget presentation choices. Advertising source IDs directly also gives future personal-data schemas one capability mechanism.

## Compatibility

`personal_data` remains optional in the capability response. The new Companion integration requires `reminders` in `personal_data.sources` and writes only the grouped `reminders` source. The new Apple Reminders widget also reads only that source. Neither implements a `reminders.fridge` fallback or automatic migration.

The server retains `personal_data_reminders` and `reminders.fridge` solely so the already published Reminders, Fridge widget can continue to work. The envelope remains `personal_data_bridge_v1`.

## Scope

This is contract-only. It does not include the server adapter, Companion EventKit/UI changes, or the new community widget. Those remain separate dependent changes after contract review.

## Validation

- `pytest -q tests/companion/test_contract_fixtures.py` — 41 passed
- `ruff check tests/companion/test_contract_fixtures.py` — passed
- `git diff --check` — passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ All notable changes to Tesserae are recorded here. Format loosely follows
   EventKit calendar identifiers. The deprecated `reminders.fridge` source and
   its legacy feature flag remain available only for the already published
   fridge widget; new Companion and widget integrations use `reminders` only.
+  The server rejects duplicate list IDs and aggregate item overflow instead of
+  truncating them, while an empty list set remains a fresh enabled snapshot;
+  deleting the source is reserved for disabling the integration.
 
 - **Physical buttons now work on relay-paired panels** (#180). A button press
   rides the status JSON the panel already posts to its relay mailbox
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Added
 
+- **The Companion API contract now supports selected Apple Reminders lists as
+  one domain source.** Servers advertise accepted schemas through
+  `personal_data.sources`; the new strict `reminders` snapshot carries up to
+  20 named list groups and 200 incomplete items in aggregate without exposing
+  EventKit calendar identifiers. The deprecated `reminders.fridge` source and
+  its legacy feature flag remain available during the compatibility period.
+
 - **Physical buttons now work on relay-paired panels** (#180). A button press
   rides the status JSON the panel already posts to its relay mailbox
   (`button` + `button_event_id`, the same fields as the REST status body); the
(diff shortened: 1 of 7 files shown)


dmellok:
Reviewed, and the shape matches what we agreed: sources advertised directly, envelope stays v1, the generic path picks up the new enum value, fridge properly deprecated, and reusing `ReminderSnapshotItem` keeps the item privacy surface identical. Publication ids instead of EventKit calendar identifiers is a good catch. Four things to pin in the text before I approve, all wording rather than structure:

1. **Aggregate cap enforcement.** The 200-items-across-all-lists cap can't be expressed in the schema (per-list `maxItems: 200` x 20 lists validates up to 4,000), so the server enforces it. Add to the `RemindersData` description: exceeding the aggregate is `invalid_snapshot` (400), rejected rather than truncated. One error code, no new enum member.
2. **Duplicate list ids.** Nothing forbids two lists with the same `id`, which would make widget list selection ambiguous. Add: list `id` values are unique within a snapshot; duplicates are `invalid_snapshot`.
3. **Empty `lists`.** The schema allows `lists: []`. I'd keep that legal and define it: an empty snapshot means "no lists currently published", the source stays fresh, and widgets show their unavailable state. That way deselecting the last list is still an atomic PUT and the app never switches verbs; DELETE stays reserved for disabling the source. If you'd rather require `minItems: 1` and use DELETE for that case, say so and I'll take that instead, but one of the two needs to be written down.
4. **No-fallback stance.** You've dropped the `reminders.fridge` fallback from the migration path relative to the discussion sketch: a new Companion on an older server simply doesn't offer the feature. I'm fine with that, it's simpler and the compatibility window is short, but state it in the contract notes as the intended behaviour so it reads as a decision rather than an omission.

One nit: the `PersonalDataCapabilities` description says "When absent, legacy clients may use legacy feature flags", but legacy clients never read this object. Suggest: "Absent on servers that predate source advertisement; those servers signal personal-data support through feature flags only."

With those four written in, this is approvable as-is and you're clear to start the dependent slices.

dmellok (approved):
All four points and the nit are addressed, and the empty-snapshot fixture makes the deselection semantics executable rather than prose. Contract's pinned from my side; go ahead with the dependent slices in whatever order suits, the server adapter is yours as offered.

PR #183, opened 2026-08-03
Implement generic Reminders server adapter

charmmmz:
## Summary

- advertise the generic `reminders` source alongside the deprecated `reminders.fridge` source
- accept and strictly validate grouped multi-list Reminder snapshots defined by #182
- reject duplicate list publication IDs, more than 20 lists, and more than 200 aggregate items with `invalid_snapshot`
- keep `lists: []` as a fresh enabled snapshot and reserve DELETE for disabling a source
- redact expired raw personal-data values while retaining a metadata-only tombstone for `expired` status
- validate required nullable Reminder due dates as real `YYYY-MM-DD` dates

## Why

PR #182 pinned the generic Apple Reminders contract so Companion can publish several explicitly selected lists using opaque publication IDs. This PR implements the server adapter offered in the review while preserving the published legacy fridge source for existing clients and widgets. New clients discover support only through `personal_data.sources` and do not fall back to the legacy source.

## Validation

- `python -m pytest -q tests/companion` — 174 passed
- targeted contract and personal-data tests — 67 passed
- `ruff check` on the changed Python modules and tests
- `mypy --python-version 3.12 --strict app/companion_api.py app/state/personal_data_store.py`

## Privacy

The server still retains only the latest snapshot per source. Once `expires_at` is reached, Reminder titles, list names, and the raw snapshot are removed; only non-sensitive timestamps remain so status can report `expired` rather than `never synced`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,12 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Expired personal-data snapshots no longer retain raw values.** The store
+  removes list names and Reminder contents at `expires_at` while preserving a
+  metadata-only tombstone for the `expired` status. Reminder due dates are now
+  also validated as real `YYYY-MM-DD` calendar dates for both the generic and
+  deprecated source schemas.
+
 - **Revoking a remote panel now actually cuts it off.** The relay Worker's
   revoke deleted the panel's mailbox but left its token record, so a revoked
   panel kept authenticating and saw an empty-mailbox `204` it couldn't tell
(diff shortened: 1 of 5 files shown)


dmellok (approved):
Verified against the pinned contract: all four review semantics are implemented with exact-message tests, the aggregate cap rejects rather than truncates, and the tombstone is what makes the `expired` status state truthful. I checked the redaction against the published fridge widget; it already guards the missing `snapshot` key, so nothing breaks. Two notes, neither blocking: requiring `due_date` on the legacy source is a tightening of live behaviour (contract-conformant, the app always sends it, but flagging it as deliberate), and the item loop is now duplicated between the two validators; a shared helper is worth folding in next time this file is touched. Merging.

PR #185, opened 2026-08-04
Add per-placement widget update-on-change policy

charmmmz:
## Summary

Adds the Stage 1 contract and placement policy discussed in [Discussion #176](https://github.com/dmellok/tesserae/discussions/176):

- let widget manifests declare supported `updates.on_change` sources, optionally narrowed by a declared `selector_option`
- persist an independent `update_on_change` opt-in on each Grid cell and Canvas widget element
- expose the switch only for capable widgets, default it off, and reset it when the placed widget changes
- document the manifest contract and add loader, Grid, and Canvas regression coverage

## Why

The server needs an explicit widget capability and a host-owned per-placement policy before change events can safely feed the existing page-refresh machinery. Keeping the policy on the placement means the same widget can opt in on one dashboard without affecting another dashboard or another placement.

This is intentionally Stage 1 only. It does not add event delivery, semantic snapshot diffing, debounce/coalescing, expiry timers, or any new scheduling path. Sync request latency and rendering behaviour are unchanged.

## Validation

- `121 passed` across `tests/test_plugin_loader.py`, `tests/test_page_routes.py`, and `tests/test_panels.py`
- Ruff passed for the touched Python source and tests
- mypy passed for the touched source modules
- JSON schema parsing and `node --check static/panels/editor.js` passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ## [Unreleased]
 
+### Added
+
+- **Widgets can declare opt-in data-change updates.** A widget manifest may
+  expose strict `updates.on_change` source declarations, optionally narrowed
+  by a declared `selector_option`. Grid cells and Canvas widget elements now
+  persist an independent `update_on_change` policy (off by default), and their
+  editors surface the switch only for capable widgets. This stage defines the
+  placement contract only; data-change delivery remains separate.
+
 ### Changed
 
 - **The Decks page got its designed look.** Implemented from a design
--- a/app/mcp_api.py
+++ b/app/mcp_api.py
@@ -1129,6 +1129,7 @@ def _drift_conflict(page: Any) -> Response | None:
 
 def _save_mcp(page: Any) -> None:
     """Stamp the page as MCP-written and persist it."""
+    _pr._coerce_canvas_update_policy(page)
     _pr._stamp(page, "mcp")
     _pr._pages().save(page)
 
--- a/app/page_routes.py
+++ b/app/page_routes.py
@@ -1184,6 +1184,16 @@ def _apply_cell_form(cell: Cell, form: Any, panel: Any) -> Cell:
         options = _cell_options_from_form(plugin, form)
     else:
         options = {}
+    if plugin_changed:
+        # The opt-in belongs to the old widget dependency, not merely the box.
+        # A newly selected widget always starts at the contract default (off).
+        update_on_change = False
+    else:
+        update_on_change = _update_on_change_from_form(
+            form,
+            cell.update_on_change,
+            supported=bool(plugin and plugin.on_change_updates),
+        )
     return cell.model_copy(
         update={
             "plugin": new_plugin_id,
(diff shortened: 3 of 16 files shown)


dmellok (changes requested):
Thanks for this. The contract shape is right (capability on the manifest, policy on the placement), the `update_on_change_present` marker handles the unchecked-checkbox vs partial-autosave ambiguity correctly, and the loader/schema/test layering matches how the rest of the plugin surface is built. Two things to fix before merge, one optional.

**1. Enforce the capability server-side on the canvas path too.**

The grid path coerces the flag to `False` for non-capable widgets in `_update_on_change_from_form`, but `POST /panels/c/<id>/save` (and MCP `set_canvas` / `patch_canvas`) will persist `update_on_change: true` on any element, including decorations and widgets with no `updates` declaration; only the editor UI hides the toggle. It also means a widget swap done through MCP or live-sync keeps a stale opt-in, since the client-side reset in `editor.js` is the only reset on that path.

I know Stage 2's coordinator will re-check the manifest before acting, so this isn't exploitable, but the asymmetry lets canvas docs accumulate meaningless `true` bits that survive widget swaps. Cheapest fix is to coerce in `_doc_to_page` (or a validator on `Element`) mirroring the grid rule.

**2. Add a save-endpoint round-trip test for the canvas flag.**

`test_element_defaults_and_roundtrip` constructs `Element` directly against the store; nothing exercises the flag through the save endpoint the editor actually uses. A small addition to an existing panels save test would pin the persistence path and document whatever you decide for point 1.

**Optional:** `catalog_entry` re-derives the capability from the raw manifest rather than using `Plugin.on_change_updates`. If that's deliberate for `_FakePlugin` duck-typing, a short comment would stop someone from "simplifying" it later.

No version bump needed in the PR; I'll bump `pyproject.toml` + `uv lock` when it lands.

dmellok (approved):
All three points addressed. `_coerce_canvas_update_policy` at both persistence choke points (`_doc_to_page` + MCP `_save_mcp`) closes the canvas gap, including decorations and the stale-flag-after-swap case, and the endpoint round-trip tests pin it properly.

Two non-blocking observations for a future pass:

- The coercion is capability-based, not identity-based: an MCP patch swapping between two capable widgets carries the opt-in forward, while the grid and editor reset on any swap. Defensible since Stage 2 matches sources against the current manifest; a one-line comment would make the asymmetry read as intentional.
- `_coerce_canvas_update_policy` both mutates and returns; `_save_mcp` relies on the mutation, `_doc_to_page` on the return. Works today, but would become a silent no-op on the MCP path if `Page` ever turns on validate-assignment/frozen. Worth a comment or a copy-and-return when you next touch it.

Merging once the pytest jobs finish. Version bump lands on main separately per repo convention.

dmellok:
Merged, squashed onto main with all checks green. The version bump lands on main separately per repo convention.

This gives Stage 2 a clean base: the coordinator can trust that any stored `update_on_change: true` sits on a placement whose widget currently declares a matching source. When you pick up event delivery, the two notes from the approval (capability-vs-identity swap semantics, and the mutate-and-return shape of `_coerce_canvas_update_policy`) are the only loose threads to keep in view.

PR #188, opened 2026-08-04
feat(widgets): refresh opted-in pages on data changes

charmmmz:
## Summary

- emit non-sensitive semantic events after accepted personal-data PUT and DELETE requests
- debounce change bursts for a 10-second quiet window, then resolve only opted-in widget placements against their current manifest declarations
- refresh active matching pages through the existing quiet-hours-aware push path and warm inactive Deck pages without promoting them
- narrow Apple Reminders changes to affected opaque `list_id` values while ignoring generated/expiry timestamps and ordering-only republishes

This implements Stage 2 of the change-driven widget refresh design discussed in #176, on top of the per-placement policy and manifest contract merged in #185.

## Contract and behavior

- personal-data storage and response semantics remain synchronous and independent from rendering or delivery outcomes
- refresh coordination is fire-and-forget after storage; enqueue, render, and push failures do not turn an accepted upload into a client retry
- DELETE invalidates all opted-in placements for its source; TTL expiry does not schedule a refresh in v1
- matching Grid and Canvas placements are deduplicated to one refresh per page
- a page shared by multiple Decks is warmed only once per device while all Deck warm timestamps are retained
- active refresh History entries use the `data_change` source

## Validation

- `python -m pytest -q tests/test_data_change_refresh.py tests/test_page_refresh.py tests/test_deck_refresh.py tests/test_deck_prerender.py tests/companion/test_companion_personal_data.py tests/companion/test_contract_fixtures.py tests/test_plugin_loader.py tests/test_page_routes.py tests/test_panels.py` — 203 passed
- `ruff check .`
- `ruff format --check .` — 463 files already formatted
- strict type check of `app/data_change_refresh.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,12 @@ All notable changes to Tesserae are recorded here. Format loosely follows
   expose strict `updates.on_change` source declarations, optionally narrowed
   by a declared `selector_option`. Grid cells and Canvas widget elements now
   persist an independent `update_on_change` policy (off by default), and their
-  editors surface the switch only for capable widgets. This stage defines the
-  placement contract only; data-change delivery remains separate.
+  editors surface the switch only for capable widgets. Accepted semantic data
+  changes are quietly debounced for 10 seconds, then active matching pages
+  refresh through the existing quiet-hours-aware push path while inactive Deck
+  pages are re-warmed without being promoted. TTL-only republishes do not
+  refresh, and personal-data PUT/DELETE responses remain independent from
+  background render or delivery outcomes.
 
 ## [0.259.0], 2026-08-04
 
--- a/app/app_factory.py
+++ b/app/app_factory.py
@@ -927,6 +927,14 @@ def create_app(
         condition_evaluator=condition_evaluator,
     )
     app.config["SCHEDULER"] = scheduler
+    from app.data_change_refresh import DataChangeRefreshCoordinator
+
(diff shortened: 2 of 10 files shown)


dmellok:
Reviewed against main and the #185 contract; this matches the Stage 2 design from #176. I checked the scheduler integration specifically: the showing-map, quiet-hours push, and the deck warm key format all line up with the existing periodic passes, and the enqueue-failure and idempotent-DELETE tests cover the response-semantics guarantees I cared most about.

Two small things, neither blocking:

1. `reminders.fridge` falls through to the generic `previous == current` comparison, which is order-sensitive on `data.items`, so an ordering-only republish fires a source-wide event. Debounce and no-change push handling absorb it, but since `_canonical_items` already exists it seems cheap to give fridge the same order-insensitive treatment.
2. The debounce restarts on every event with no max-latency cap, so a sustained sub-10s publish stream defers the refresh indefinitely. Fine as a v1 behaviour if the Companion publish cadence makes it unlikely; flagging so it's a decision rather than an accident.

Either fold the fridge canonicalization into this PR or mark it ready for review as is and I'll merge, with that as a follow-up. Your call on 2 as well; a one-line comment documenting it as intentional would be enough.

dmellok:
Merged. CI came back green across the board (full pytest suite plus the 3.12/3.13 forward-compat runs), and the fridge canonicalization + debounce comment cover both review points. This will go out in the next edge version.

PR #189, opened 2026-08-05
fix: isolate personal data publishers

charmmmz:
## What changed

- key personal-data snapshots by paired Companion installation and source instead of source alone
- scope PUT ordering, status, and DELETE to the publisher identity bound to the authenticated pairing
- expose all publications to server-side widgets with the paired client name as display metadata
- migrate the previous single-publisher JSON shape in memory and retire each legacy source on its first authenticated sync
- narrow first-sync and delete refresh events to the affected Reminders list IDs when they are known
- document the unchanged client payload contract and the trusted installation-descriptor boundary

## Why

The previous store used only `source_id` as its key. In a household with two paired Companion apps, the second phone could replace the first phone’s Reminders snapshot, an older snapshot from the second phone could be rejected by the first phone’s timestamp, and either phone could delete the shared record when disabling sync.

Publisher identity is derived from the stable installation descriptor already bound to the authenticated pairing. No publisher or user field is added to the snapshot payload. Installation descriptors are trusted within the Companion pairing boundary; independently identified installations receive separate publications, while re-pairing the same app installation continues to address one publication.

## User impact

Independently identified household phones can publish the same personal-data source without replacing each other. Each phone sees the freshness status and controls the publication associated with its pairing. Multi-publisher-aware widgets can enumerate every publication and distinguish duplicate list titles using the paired client name. Existing widgets retain a newest-generated-publication fallback until updated.

## Validation

- full `python -m pytest -q` before the upstream-only rebase: 2919 passed
- post-rebase focused personal-data, contract, and change-refresh tests: 68 passed
- Ruff format/check on changed Python files: passed
- strict mypy on `personal_data_store.py` and `data_change_refresh.py`: passed
- Apple Reminders widget compatibility smoke tests against this branch: 11 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [Unreleased]
+
+### Fixed
+
+- **Personal-data snapshots are isolated per paired Companion installation.**
+  Two household phones can publish Apple Reminders without overwriting each
+  other, each phone sees only its own sync status, and disabling sync cannot
+  delete another phone's publication. Existing single-publisher data remains
+  readable and is replaced by the first authenticated publisher that next
+  syncs that source.
+
 ## [0.266.0], 2026-08-05
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 ### Fixed
 
 - **Personal-data snapshots are isolated per paired Companion installation.**
-  Two household phones can publish Apple Reminders without overwriting each
-  other, each phone sees only its own sync status, and disabling sync cannot
-  delete another phone's publication. Existing single-publisher data remains
-  readable and is replaced by the first authenticated publisher that next
-  syncs that source.
+  Independently identified household phones can publish Apple Reminders
+  without replacing each other, each phone sees only its own sync status, and
+  disabling sync deletes only the publication associated with its pairing.
+  Existing single-publisher data remains readable and is replaced by the first
+  authenticated publisher that next syncs that source.
 
 ## [0.266.0], 2026-08-05
 
--- a/app/companion_api.py
+++ b/app/companion_api.py
@@ -25,6 +25,7 @@ mypy --strict applies to this module, see pyproject.toml.
 
 from __future__ import annotations
 
+import hashlib
 import json as _json
 import logging
 import re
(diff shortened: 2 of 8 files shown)


dmellok:
The publisher-scoped keying and the lazy legacy migration both look right, and the test coverage of the isolation scenarios is solid. A few things before this lands:

**Contract wording overstates the guarantee.** Publisher identity is a hash of the client-supplied `installation_id`, so a second device could pair claiming another phone's installation id and address its publication. That's acceptable inside the pairing trust boundary, but the new DELETE description says it "cannot delete another paired installation's snapshot", which reads as a security property we don't actually have. Could you soften it to something like "another installation's snapshot is not addressable through this API" or note the honest-client assumption?

**Newest-publication fallback ordering.** The no-publisher `get()` picks the winner by `stored_at`. A phone that syncs frequently but with an older `generated_at` (clock skew, delayed background snapshot) would win over semantically newer data, which is the same failure mode the ordering check exists to prevent. Suggest max `generated_epoch` instead. Also there's no test exercising this fallback with two live publishers, and it's the exact behavior existing widgets depend on until they adopt `publications()`; one test there would be good.

**Silent legacy default on `put()`/`delete()`.** Both in-repo call sites now pass `publisher_id` explicitly. A future caller that forgets the kwarg writes into the legacy namespace and gets its data retired on the next companion sync, silently. Since the tests all pass it explicitly, make `publisher_id` keyword-required with no default and keep the legacy constant internal to migration.

**Orphan cleanup (follow-up, not blocking).** A revoked or reinstalled phone's publication lingers as a fresh record until TTL and then a tombstone forever, with no owner able to delete it. Worth a follow-up issue: drop a publisher's publications on token revocation, or purge tombstoned publishers after some horizon.

**Post-upgrade status.** After migration an existing phone's `GET /personal-data/status` returns `sources: []` until its next PUT. Can you confirm the app treats empty status as "re-sync" rather than rendering "never synced" persistently?

Housekeeping I'll handle at merge: version bump + `uv lock`, and renaming the `[Unreleased]` CHANGELOG heading to the real version.

charmmmz:
Addressed the blocking feedback in `593c6c4` and rebased onto current `main`:

- softened the contract and changelog language to state the trusted installation-descriptor boundary rather than claim an absolute cross-client security property
- changed the compatibility `get()` fallback to choose the maximum `generated_epoch`, with a two-live-publisher regression test where the later-stored snapshot is semantically older
- made `publisher_id` a required keyword for `put()` and `delete()` so new callers cannot silently write into the migration-only legacy namespace

Post-upgrade iOS behavior is also safe: an enabled bridge loads an empty status as `sourceStatus == nil`; the root view then starts active change monitoring, and the automatic sync path treats nil status as requiring upload even when the content digest is unchanged (`automaticUploadIsRequired` returns true unless status is fresh). The UI shows `On` during that short window rather than persisting a “never synced” state. This nil-status behavior already has unit coverage in `testAutomaticUploadDecisionRequiresChangeOrNonFreshServerSnapshot`.

Post-rebase validation: 68 focused server/contract tests passed, strict mypy and Ruff passed, and the multi-publisher Apple Reminders widget has 11 passing smoke tests against this branch. I agree orphan cleanup belongs in a follow-up rather than this PR.

PR #201, opened 2026-08-08
PicPak BWRY calibration

varanu5:
# Palette calibration support for the 4-colour BWRY gamut (PicPak)

## Summary

`bwry_4` is the only supported gamut with no palette-profile family, so the
Calibration tab hides entirely for those panels and frames are always dithered
against the ideal sRGB primaries. This adds the missing family and a calibrated
palette, using the machinery Spectra 6 and Inky 7-colour already have.

Everything is scoped to `bwry_4`. No other panel changes behaviour — see
[Isolation](#isolation) for how that is enforced.

Validated on a physical PicPak 4.2" panel, not only in tests.

## Why

`BWRY_4_PALETTE` targets `(0,0,0) / (255,255,255) / (255,255,0) / (255,0,0)`.
Real BWRY ink reaches none of those. Yellow is the worst case: the quantiser
treats `(255,255,0)` as a bright highlight while the panel lays down a dark
mustard, so error diffusion spends yellow on highlights the ink cannot deliver.
The reproducible range is roughly `[36, 236]`, not `[0, 255]`, so an
uncalibrated dither also over-drives contrast against endpoints that do not
exist.

`_CALIBRATED_PALETTES` plus `_compress_to_calibrated_range` already solve this
for other gamuts, but the map only held `waveshare_e6` and `inky_7colour`, so
the calibrated path was unreachable for BWRY. Separately,
`_palette_family_for()` returned `""` for `bwry_4`, which null-gates every
palette endpoint and hides the editor.

## What changed

**`app/quantizer.py`**
- Adds `BWRY_4_CALIBRATED_PALETTE` and registers it in `_CALIBRATED_PALETTES`.
  Measured against a physical PicPak 4.2" panel by
  [varanu5](https://github.com/varanu5); epdoptimize carries no BWRY profile, so
  these are an independent contribution rather than a port.

**`app/palette_profiles/bundled.py`**
- Adds a `bwry_4` family with two presets:

  | preset | slug | purpose |
  |---|---|---|
  | **Nominal BWRY** | `nominal-bwry` | **family default.** Ideal primaries — byte-identical to having no profile applied. |
  | **PicPak Calibrated** | `picpak-bwry-calibrated` | Measured palette. Carries `based_on` so the picker card credits the source. |
- `_profile()` gains an optional `dither=` argument. Every pre-existing preset
  omits it and constructs exactly as before.

**`app/settings/index_routes.py`**
- `_palette_family_for()` maps `bwry_4` to its own family, after the existing
  Inky and Spectra branches.
- The swatch grid drops blue/green for BWRY. `PaletteColors` keeps six slots so
  the stored schema is unchanged for every family, but the quantiser slices a
  profile palette to the gamut's length, so those two would be controls that
  silently do nothing.
- Adds `bwry_4` rows to the test-pattern label/hex maps, so the solid-fill
  picker stops offering a BWRY panel blue and green it cannot paint.

**`app/test_patterns.py`**
- `_palette_for()` knew `inky_7colour` and fell through to the six-colour E6
  deck for everything else, `bwry_4` included. Three failures followed from that
  one gap, all visible when sending **Palette swatches** to a real panel:
  - the pattern painted blue and green blocks the panel has no ink for;
  - `build_pattern` discards an override shorter than the deck
    (`len(override) >= len(palette)`), so a 4-entry BWRY profile never reached
    the pattern and the user's chosen profile had no effect on what was sent;
  - Floyd-Steinberg diffuses error rightward, so the out-of-gamut yellow column
    bled into the red column and speckled it.

**`app/settings/devices_routes.py`**
- The live-preview swatch threshold was a hardcoded `6`; BWRY posts four.

**`NOTICES.md`** — attribution section for the BWRY calibration data.

## Design notes

**The default is deliberately the *uncalibrated* profile.** This inverts the
convention used by the other families, where the default is measured and Nominal
sits last as the "start over" option. The reason: `_palette_profile_slug_for()`
self-heals a blank slug to the family default and persists it, so shipping a
calibrated default would silently restyle every existing BWRY panel on next
render. `nominal-bwry` renders byte-identical to no profile at all, which a test
asserts.

Making that true required pinning `serpentine=False` on the presets —
`DitherSettings` defaults it to `True` while the renderer's no-profile path
defaults it to `False`, so without the pin merely *having* a profile applied
would change output.

If you would rather BWRY follow the same convention as the other families,
reorder the presets in `bundled.py` and change `default_slug_for()`; the
docstring records this.

## Known characteristics, documented rather than worked around

**Light red speckle on neutral text under the calibrated palette.** The real red
ink is desaturated enough to sit next to neutral mid-grey in colour space, so
grey levels ~80–160 snap to red — and antialiased text edges live in that band.
Roughly 0.4% of pixels on a text-heavy frame. A preset with a de-neutralised red
removed it entirely but was dropped as unwanted; **Nominal BWRY** stays clean
here and remains the right choice for text-heavy dashboards.

**Text softening on yellow / red backgrounds** (black pixels 293 → 228 on a
200×60 text scene). That is `_compress_to_calibrated_range` reducing source
contrast before dithering, not a palette problem. It cannot be disabled from a
profile, and fixing it means changing behaviour shared with Spectra 6 and Inky,
so it is deliberately out of scope.

**The swatch pattern is a poor instrument for judging a calibrated profile.**
The swatch is filled with the ink's own value, but the compression pass then
maps it off that value (`212,36,42 → 202,65,65`), so it can never land cleanly
and will always dither (red 78% vs Nominal's 100% flat). Judge on real content.

## Isolation

Every gamut-keyed addition is additive (new dict keys, new list entries) and
read only when the gamut matches. The one shared-code edit is the optional
`dither=` parameter on `_profile()`; every pre-existing preset omits it and
constructs identically.

Verified empirically rather than by inspection:
- All 7 pre-existing bundled profiles serialise **byte-identical** to `main`.
- `_palette_for()` compared across 14 gamuts × 2 calibrated settings — `bwry_4`
  is the only one that differs; unknown gamuts still fall back to E6.
- `waveshare_e6`, `inky_7colour`, `bwr_3`, `mono` and `gray_4` render
  byte-identical with the new `_CALIBRATED_PALETTES` entry present versus
  monkeypatched out, at both `calibrated` settings.
- Uncalibrated `bwry_4` is byte-identical too, so panels without a profile are
  unaffected.
- Non-BWRY swatch grids stay six-slot, and the preview threshold expression is
  pinned by a test — an earlier revision generalised it by gamut and silently
  moved `inky_7colour` from 6 required swatches to 7.

## Upstream bug found along the way (not fixed here)

Surfaced by painting a LAB-matching preset to a real panel: blues came out solid
black and greens solid yellow, with no dither texture anywhere.

`color_match="lab"` silently disables dithering. In `_error_diffusion`, the
`buf_l` / `buf_a` / `buf_b_lab` buffers are built once from the original image
and never receive diffused error — only `buf_r/g/b` do — yet the LAB branch
picks the nearest ink by reading those stale buffers. Every pixel is judged
against its pristine value, so error diffusion is ignored entirely and the
output is plain nearest-colour quantisation.

Signature, on a flat 40×40 mid-grey patch:

```
color_match=rgb  ->  white 48%, black 47%, red 3%
color_match=lab  ->  white 100%
```

**This is not BWRY-specific, and not Floyd-Steinberg-specific.**
`_error_diffusion` never sees a gamut, and `atkinson` / `jarvis` / `stucki` all
call it directly (`floyd-steinberg` detours into it when LAB is selected).
Measured across every combination — a flat patch collapses to **one ink in
12/12** gamut × dither pairings, under both `lab` and `chroma-aware`, where RGB
uses 4 to 7:

| gamut | rgb | lab / chroma-aware |
|---|---|---|
| `bwry_4` (4 inks) | 4 inks | **1** |
| `waveshare_e6` (6 inks) | 6 inks | **1** |
| `inky_7colour` (7 inks) | 5–7 inks | **1** |

So a Spectra 6 or Inky user who selects LAB matching in the Calibration tab —
which is offered as a live control — gets a posterised frame with no dither
texture at all.

Fixing it means recomputing LAB per pixel after each error write, inside a
Python hot loop shared by every panel, so it is out of scope for this PR. No
preset here selects LAB matching, and `tests/test_bwry_calibration.py` carries a
regression guard across all four dithers and both LAB modes that will fail once
the bug is fixed — at which point a LAB preset becomes worth adding.

Recorded here because it is the reason this PR ships no LAB preset, and because
the investigation turned it up. Nothing in this PR depends on it being fixed —
lift it into its own issue if that suits your tracking better.

## Testing

- `tests/test_bwry_calibration.py` — 38 tests covering the palette, the profile
  family, the test-pattern deck, the UI surfaces and the isolation guarantees.
- Two existing tests used `bwry_4` as their stand-in for "a gamut with no
  profile family", a premise this PR reverses. They now use `mono`, which is
  what one of them already said in its own comment. Their assertions are
  otherwise unchanged.
- 710 passing across the palette / quantiser / render / settings / device
  suites. `ruff` clean.

## Hardware validation

Tested visually on a real PicPak 4.2" BWRY panel.

- **Nominal BWRY** — byte-identical to no profile applied (asserted by test),
  i.e. the output this panel has always produced.
- **PicPak Calibrated** — painted to the panel. The red speckle on antialiased
  text and the softening on yellow/red backgrounds were both **observed on glass
  first**; the numbers quoted above were derived afterwards to explain them.
- The palette-swatches bug above was found the same way — by sending the pattern
  to the panel and noticing blue and green blocks it cannot print.
- A LAB-matching preset was built, painted, and **rejected on the panel**, which
  is what led to the upstream bug.

Caveat: the values come from one panel, and batches vary. The profile notes
point users at forking the preset and adjusting the swatches.

### DIFF
--- a/NOTICES.md
+++ b/NOTICES.md
@@ -48,3 +48,21 @@ the calibration data. The presets live in
 [`app/palette_profiles/bundled.py`](app/palette_profiles/bundled.py);
 user-authored profiles are saved to
 `data/palette_profiles/<slug>.json` and follow the same schema.
+
+## varanu5 — PicPak 4.2" BWRY calibration
+
+The 4-colour BWRY calibrated palette (in
+[app/quantizer.py](app/quantizer.py) as `BWRY_4_CALIBRATED_PALETTE`, and
+shipped as the `picpak-bwry-calibrated` preset in
+[`app/palette_profiles/bundled.py`](app/palette_profiles/bundled.py)) was
+measured against a physical PicPak 4.2" black/white/red/yellow panel by
+[varanu5](https://github.com/varanu5), author of the PicPak community
+firmware. epdoptimize carries no BWRY profile, so these values are an
+independent contribution rather than a port.
+
+The preset carries `based_on` and `attribution` fields that surface as a
+"via varanu5" chip on the picker card, the same way the
+paperlesspaper-derived presets do.
+
+* Upstream: https://github.com/varanu5
+* PicPak firmware: https://github.com/varanu5/picpak-tesserae-client
--- a/app/palette_profiles/bundled.py
+++ b/app/palette_profiles/bundled.py
@@ -258,10 +258,10 @@ BUNDLED_PROFILES: tuple[PaletteProfile, ...] = (
         # the card renders it without a link. ``attribution`` exists to point
         # at an upstream project the data was ported from, which does not
         # apply here.
-        based_on="varanu5 · PicPak 4.2\" BWRY panel calibration",
+        based_on='varanu5 · PicPak 4.2" BWRY panel calibration',
         notes=(
             "The colours this panel actually prints, measured from a real PicPak "
-            "4.2\" by varanu5. The default profile aims for perfect ink the screen "
+            '4.2" by varanu5. The default profile aims for perfect ink the screen '
             "can't make (its yellow is closer to mustard than lemon), which can "
             "leave photos looking muddy. Panels vary a little, so copy this "
             "profile and adjust the colours if yours looks off."
--- a/app/palette_profiles/bundled.py
+++ b/app/palette_profiles/bundled.py
@@ -1,17 +1,27 @@
 """Read-only palette presets shipped in the source tree.
 
-Six Spectra 6 profiles + one seven-colour Inky ACeP profile + one
-nominal (uncalibrated identity) fallback. Every measured profile is
-attributed to its upstream source; the palette hex values for the
-paperlesspaper-attributed profiles are the Apache-2.0 palette data
-from `paperlesspaper/epdoptimize
+Grouped by panel family: Spectra 6, seven-colour Inky ACeP, and 4-colour
+BWRY, each with at least one measured profile plus a nominal
+(uncalibrated identity) fallback. Every profile derived from someone
+else's work is attributed to its upstream source; the palette hex values
+for the paperlesspaper-attributed profiles are the Apache-2.0 palette
+data from `paperlesspaper/epdoptimize
 <https://github.com/paperlesspaper/epdoptimize/blob/main/src/dither/data/default-palettes.json>`_.
+The BWRY values were calibrated against a physical PicPak 4.2" panel by
+`varanu5 <https://github.com/varanu5>`_.
 
 Contributors ordered so the picker's default is Tesserae's tuned
 profile (currently equivalent to paperlesspaper's ``spectra6``), then
 the community-measured alternatives in rough popularity order, then
 Nominal at the bottom as the "start over" option. The order the
 picker renders in comes from :func:`list_bundled`.
+
+**BWRY deliberately inverts that**: its nominal profile is listed first
+and is the family default, so unlocking the Calibration tab for PicPak
+panels does not silently restyle frames that already look right. The
+measured presets are one click away. If BWRY should follow the same
+convention as the other families, reorder here and change
+:func:`default_slug_for`.
 
 Attribution surfaces:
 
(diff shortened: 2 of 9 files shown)


dmellok:
Merged, on edge in v0.282.0.

The isolation work is what made this quick to take, and measuring the panel first and deriving the numbers afterwards is the right way round.

Two small things I folded into a follow-up rather than sending back: the per-device "Calibrated palette" toggle now actually does something on a PicPak, which it hasn't until now, so I've pinned that with a test; and a stale docstring.

The LAB dithering bug is the most valuable part of this and I'll open it as its own issue.

PR #210, opened 2026-08-10
Widget/calendar native refresh

tonyjamesstark:
Repo: tesserae
Branch: widget/calendar-native-refresh

Title: Calendar widgets: configurable text sizing, rolling 7-day week option, multi-day all-day fix

## Summary
Numerous fixes to allow more configuration to calendar widgets. Authored with Claude

Adds a `range_mode` option to calendar_week for a rolling 7-day window
("today + next 6" instead of the fixed calendar week), with per-day
weekday now driving labels/shading so it's correct in both modes. Fixes a
bug where calendar_week bucketed multi-day all-day events (e.g. a 3-day
conference) only under their start date instead of every day they span —
calendar_month already had the correct pattern, applied here — and adds
the missing all-day pill row to calendar_week's client, which had the CSS
support but never rendered it. Also folds in general cleanup: reordered
cell_options groupings, removed stale ImportError fallbacks in
calendar_day, widened calendar_month's padding slider bounds to match its
client clamp. 

Audited but unchanged: calendar_day already handled multi-day all-day
events correctly.

**Follow-up (7d7fd441, 4d43d736):** the day-bucketing fix above only
solved half the multi-day problem — client.js rendered a *separate* pill
in every covered day's column with the label repeated, instead of one
bar spanning them. `4d43d736` groups the per-day copies back by (start,
end, summary) and renders a single pill positioned with
`grid-column: start / span N`, lane-packed so overlapping all-day events
stack instead of colliding.  Also `7d7fd441`: the shared `slider` cell-option macro
(`_components.html`/`components.js`/`forms.css`) gets an editable number
input alongside the range input — one change point that covers every
`type: slider` option app-wide, including the calendar text-size scales.

## Files touched
CHANGELOG.md, plugins/calendar_day/{client.js,server.py,tests/{test_smoke.py,clamp_check.mjs}},
plugins/calendar_month/{client.js,plugin.json},
plugins/calendar_week/{client.js,plugin.json,server.py,tests/{test_smoke.py,clamp_check.mjs}},
plugins/calendar_core/server.py, tests/test_calendar_core_caldav.py,
app/calendar_time.py,
templates/_components.html, static/components.js, static/style/forms.css

**Follow-up (2eb2c8d5, 2418a135, c4d9c6a3, 7acd7419, 460d3860):** a second
round of multi-day-event bugs, found via a device-preview screenshot
after the fixes above landed:

- A timed event whose end lands on a later local day than its start (an
  overnight block, a multi-day trip) subtracted raw hour-of-day across
  two different dates, producing a negative/near-zero height and
  corrupting `computeRange`'s auto-fit window for the whole visible
  day/week (`2eb2c8d5`, calendar_day + calendar_week).
- `calendar_time.event_local_date_key` only ever returned an event's
  *start* day, so calendar_week bucketed a multi-day *timed* event on its
  first day only — unlike all-day events, it never appeared on any day
  after that. Added `timed_event_date_keys` mirroring the existing
  all-day spread (`2418a135`).
- Client-side, every per-day copy of a multi-day timed event redrew at
  the event's original absolute start hour instead of clamping to the
  day actually being rendered, and always ran to midnight even on the
  real end day. Added `clampToDay` in calendar_week (`2418a135`) and
  calendar_day (`7acd7419`) to clamp each day-copy against that day.
- The clamp above could still push a block's height past 100% and spill
  past the lane's bottom edge when `day_start_hour`/`day_end_hour`
  narrowed the visible range below a pass-through day's full 0-24h span.
  Added `pctSpan` to clamp top/bottom independently in calendar_week
  (`c4d9c6a3`) and calendar_day (`7acd7419`).
- All-day pills were shifted off their true day/lane column: in
  calendar_week the nested per-week grid declared an unsized `auto`
  gutter column that collapsed to 0 (`2418a135`); in calendar_day the
  pill row was a full-width sibling of the hour-gutter+lane grid
  entirely, offset by the gutter's width (`460d3860`). Both now share
  the real grid's column tracks instead of guessing at a matching width.

**Follow-up (99ef72d2):** a third bug family, reported directly ("the day
view now skips all-day events") rather than found by audit. Root cause was
in `calendar_core.load_events`'s shared warm-cache slicing
(`_expand_events_cached`), not any individual widget: it compared an
all-day event's bare `"YYYY-MM-DD"` date string against the query
window's full UTC datetime string. A bare date string sorts as "less
than" any same-day timestamp string, so once the UTC calendar day rolled
past local midnight — any evening in a negative-UTC-offset timezone, e.g.
~7pm US Eastern — today's all-day event silently dropped out before ever
reaching each widget's own correct local-date filtering
(`all_day_event_overlaps_date`/`all_day_event_date_keys`). Widened the
window by a day on each side for all-day events instead of tightening the
string comparison; every widget already re-filters by real local date
downstream, so over-inclusion here is harmless. One fix in `calendar_core`
covers every widget that calls `load_events` — bundled and standalone
alike, including `tesserae-widget-calendar-three-day` and
`tesserae-widget-calendar-schedule`, with no separate per-widget fix
needed.

**Follow-up (5ee12e75):** a fourth bug, same symptom report as the third
("day view all-day events") but a different root cause the string-comparison
fix above didn't touch: once events *were* in the response, calendar_day's
all-day pills rendered only as wide as their own titles, stopping short of
the event column's right edge — an all-day event read as a small tag rather
than a bar covering the day it applies to. The shared `.tt-allday` in
spectra-widgets.css is a wrapping flex *row*, and its pills are flex items,
so they size to content. calendar_week and calendar_three_day never hit
this because they don't use `.tt-allday` at all — they build their own
`.tt-allday-row` grid and span each pill with an explicit
`grid-column: N / span M`. Day view has exactly one lane, so the fix is one
declaration: stack the pills vertically (`flex-direction: column`) and let
them stretch the lane end to end.

## Test plan
- `uv run pytest plugins/calendar_day plugins/calendar_week
  plugins/calendar_month` — all passing, including new regression tests
  `test_multi_day_all_day_event_appears_on_every_covered_day` and
  `test_multi_day_timed_event_appears_on_every_covered_day` in
  calendar_week.
- `uv run pytest tests/test_calendar_core_caldav.py` — passing, including
  new regression test
  `test_expand_events_cached_keeps_all_day_event_past_utc_midnight`,
  which seeds the warm cache directly and slices it with a window that
  reproduces the UTC-rollover drop.
- `node plugins/calendar_day/tests/clamp_check.mjs`,
  `node plugins/calendar_week/tests/clamp_check.mjs` — `clampToDay`/
  `pctSpan` self-checks (start/pass-through/end day, full-range and
  narrowed-range overflow clamping).
- **Not covered by any automated check:** the all-day pill full-width fix
  (`5ee12e75`) is a single CSS declaration whose effect only exists once a
  layout engine resolves it. 
- Note: a pre-existing test-isolation bug (non-namespaced `import server`
  colliding across plugin test dirs when the *entire* suite runs in one
  pytest session) predates this branch and is out of scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,16 @@ All notable changes to Tesserae are recorded here. Format loosely follows
   always accepted anything up to a day, but a value the dropdown didn't offer
   had no matching option and the list drew it as "only when pushed", so opening
   the dropdown and picking anything silently rewrote it.
+- **All-day events no longer disappear from calendar_day/week/month/schedule
+  in the evening.** Every calendar widget's shared event-fetching layer
+  compared an all-day event's date against the requested time window as if
+  both were full timestamps; once the UTC calendar day rolled past local
+  midnight — any evening in a timezone behind UTC, such as US timezones —
+  today's all-day events silently dropped out of the response.
+- **calendar_day's all-day events span the full width of the event column.**
+  They were laid out in a wrapping flex row, so each pill was only as wide as
+  its own title and stopped short of the column's right edge, reading as a
+  small tag instead of a bar covering the day it applies to.
 
 ## [0.283.0], 2026-08-09
 
@@ -439,6 +449,17 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Multi-day events now display correctly in calendar_day and
+  calendar_week.** A multi-day all-day event (e.g. a 3-day conference)
+  previously appeared only on its first day, or as a separately-labeled
+  pill repeated on each day instead of one bar spanning them. A
+  multi-day *timed* event (an overnight block, a multi-day trip) is a
+  more common case and had it worse: it vanished entirely after its
+  first day, could render at the wrong time or run past midnight on
+  later days, and could overflow the visible timetable when
+  `day_start_hour`/`day_end_hour` narrowed it. All-day pills in both
+  widgets could also draw shifted left of their true day column. Every
+  day an event covers now renders correctly clamped to that day.
 - **Personal-data snapshots are isolated per paired Companion installation.**
   Independently identified household phones can publish Apple Reminders
   without replacing each other, each phone sees only its own sync status, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -424,6 +424,19 @@ All notable changes to Tesserae are recorded here. Format loosely follows
   already-registered device. The client protocol doc now spells out the
   attribute's behaviour on both `/discover` and `/register`.
 
+### Added
+
+- **Configurable text sizing for calendar_day, calendar_week, and
+  calendar_month.** Per-text-type scale sliders (title, header, axis
+  label, event title/location/time, 0.01–10.0×), a `date_label_style`
+  choice with a full-word option, and a `show_location` toggle for
+  the widgets that lacked one. calendar_day also gets configurable
+  `day_start_hour`/`day_end_hour` and a `show_density` toggle;
+  calendar_week gets a `range_mode` option so its 7-day window can
+  either stay anchored to the chosen week-start day or roll from
+  today (`rolling_7`), and correctly labels/shades each day by its
+  actual weekday rather than its column position in either mode.
+
 ### Fixed
 
 - **Personal-data snapshots are isolated per paired Companion installation.**
--- a/app/calendar_time.py
+++ b/app/calendar_time.py
@@ -11,6 +11,18 @@ def local_midnight_utc(day: date, zone: tzinfo) -> datetime:
     return datetime.combine(day, time.min, tzinfo=zone).astimezone(UTC)
 
 
+def _local_datetime(raw: str, zone: tzinfo) -> datetime | None:
+    if not raw:
+        return None
+    try:
+        parsed = datetime.fromisoformat(raw)
+    except ValueError:
+        return None
+    if parsed.tzinfo is None:
+        parsed = parsed.replace(tzinfo=UTC)
+    return parsed.astimezone(zone)
+
+
 def event_local_date_key(event: dict[str, Any], zone: tzinfo) -> str:
     """Bucket a calendar event by the date a user sees in ``zone``."""
     start = str(event.get("start") or "")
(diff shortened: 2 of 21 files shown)


dmellok:
Reviewed and pushed the CI fixes to your branch as `76899fa`. Three real bug fixes and a genuinely useful option, so this is good to go once it's green.

Notes on what was failing, since one of them is worth knowing about for future widget tests.

**The three `KeyError: 'start'` failures were a module-name collision.** The tests did `import server`, but every `calendar_*` widget ships a `server.py`, so under a full run that name resolved to whichever plugin's tests loaded first, and calendar_week's assertions ran against calendar_day's `fetch()`. Running the file on its own hides it completely. They now load under a plugin-specific module name via `importlib`, which also leaves calendar_day's existing test untouched.

**The fourth failed even in isolation, and it was the test rather than the feature.** It wrote naive event times (`...T16:00:00`) while the window came from `app_timezone()`, which falls back to the host's zone. On a UTC runner that agrees with itself; on my machine (UTC+10) the day-0 event landed on day 1. `calendar_core` normalises timed events to UTC ISO before any widget sees them, so I pinned the zone in the test and gave the fixture the offset production actually emits. `timed_event_date_keys` was correct as written.

**Two fixes in the shared slider macro**, since that part touches every widget rather than just the calendars:

- The number box wrote straight through to the range input, which clamps silently. Typing `9` into a 0.7–1.5 slider left `9` on screen while `1.5` was submitted. It now reflects back the value that will actually be saved, on change and blur.
- Added an `aria-label`; the visible `<label>` points at the range input, so the number box had no accessible name.

The design is sound otherwise: the number input stays unnamed so the range keeps carrying the field value, and its events still bubble to the form, so the editor's dirty-tracking and live preview fire as before.

**One follow-up, deliberately not in this PR.** `calendar_month` still buckets multi-day *timed* events on their start day only (`plugins/calendar_month/server.py:69`), which is the same bug this fixes for calendar_week, and the shared helper now exists. I left it because making a 3-day trip occupy three month-grid cells is a visible change for existing users and deserves its own commit and test rather than a silent ride-along. Happy to take it separately if you want it.

PR #217, opened 2026-08-11
Fix unmatched dynamic select values

charmmmz:
## What this changes

Render an explicit selected option when a saved select value is absent from
its current choices. Empty values remain visibly unset, while removed values
remain visible as unavailable until the user deliberately replaces them.

This shared component covers runtime-backed Gallery folder, Home Assistant
entity, Todo list, grid Dashboard, and Canvas configuration pickers.

## Why

Browsers visually select the first `<option>` when no option matches a
`<select>`'s saved value. A newly selected Gallery widget could therefore store
`folder=""` (root) while displaying `family (5)`. The preview correctly rendered
root and reported it empty, but the editor looked as though `family` was already
configured. An unrelated later save could also submit that browser-selected
first choice and silently change the configuration.

Keeping the real unmatched value as an explicit option makes the UI and
submitted form agree without automatically choosing a potentially destructive
first entity, folder, or list.

## Test plan

- [x] `pytest -q tests/test_page_routes.py` — 81 passed
- [x] `pytest -q tests/test_panels.py` — 42 passed
- [ ] `mypy app` — local reused Python 3.12 environment has NumPy stubs using
  Python 3.12-only `type` syntax while this repository configures mypy for
  Python 3.11; CI remains the authoritative standard-environment check
- [ ] Manually verified the patched UI in a deployed browser; the production
  instance remains unmodified

## Checklist

- [x] Tests added for the new behaviour
- [ ] `ruff` + `mypy` clean — Ruff passed; mypy was blocked by the local
  dependency/environment mismatch described above
- [x] User-facing change documented in `CHANGELOG.md`
- [x] Version bump not applicable; this PR does not publish a release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [Unreleased]
+
+### Fixed
+
+- **Dynamic widget pickers no longer look configured when their saved value is
+  empty or unavailable.** Gallery folders, Home Assistant entities, Todo lists,
+  and other runtime-backed selects now show the actual unset or removed value
+  until it is deliberately replaced, instead of letting the browser display
+  and later submit the first available option.
+
 ## [0.292.0], 2026-08-11
 
 ### Added
--- a/templates/_components.html
+++ b/templates/_components.html
@@ -139,12 +139,28 @@ macro here instead.
     ``group`` are wrapped in a single <optgroup> labelled with the group
     name, so callers can keep the picker single-element while users see
     related themes / fonts / etc. clustered. -#}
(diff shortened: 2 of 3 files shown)


dmellok:
Landed on main in v0.292.2 as fa65443, your commit kept as-is.

Two follow-ups on top in 691356d:

- `select_field` compared the saved value to option values with a bare `==`, so a string value against an integer choice (the kind-default sleep intervals) missed the match and rendered an "Unavailable" duplicate of an option already in the list. Both the scan and the option loop now compare stringified values.
- The caller's blank option used `not value` while the new synthetic option used `value != ''`, so a value of `0` came out with two options selected and browsers honour the last one. They share one emptiness test now.

Also added cases for `blank_label` + unmatched value, `blank_label` + unset, and the tuple-options form.

On scope: because the `blank_label is none` branch fires for empty values too, static pickers gain a `(not set)` entry they didn't have, e.g. Panel size on the Add device form. The JS sets that select on load and the server falls back to manifest defaults, so it isn't reachable in practice, and narrowing the condition would bring the Gallery case back. Left as-is.

PR #218, opened 2026-08-11
feat: add placement-level scheduled widget refresh core

charmmmz:
## Summary

- add the manifest updates.on_schedule contract, initially limited to daily triggers
- persist optional daily schedules per Grid cell and Canvas widget placement
- resolve scheduled placements independently from data-change dependencies
- share daily due-time semantics with existing Schedules
- coalesce due placements per dashboard, respect quiet hours without logging a skip every scheduler tick, and warm inactive Lineup pages without promoting them
- apply a five-minute retry backoff after actual scheduled refresh failures while preserving existing data-change cadence behavior

## Why

Time-dependent widget output can become stale even when its source payload has not changed. A relative due-day label is the motivating example: the date boundary changes what should render, but there is no new Reminders snapshot to produce an on-change event. Lineups decide which dashboard is visible; this contract complements them by invalidating time-dependent widget output without selecting or advancing content.

Discussion: https://github.com/dmellok/tesserae/discussions/216

## Scope

This is the Core slice only. It is off by default, supports daily schedules only, and deliberately does not add cron. The Configure UI follows in a stacked PR. A schedule without at uses the server local day boundary; suggested_at remains a manifest hint for the later UI.

Like the existing Schedules implementation, due/satisfied state is held in memory. Restarting after a day's target can defer that placement until the next day; persistence is intentionally left for a follow-up.

This is versioned as 0.293.0 and the lockfile was refreshed.

## Validation

- 49 scheduler, page-refresh, and plugin-loader tests passed
- broader stacked regression set: 279 tests passed
- Ruff passed
- mypy passed across 178 app modules with the local Python 3.12 target override
- git diff --check passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [0.293.0], 2026-08-12
+
+### Added
+
+- **Placement-level daily widget refreshes.** Widgets may declare
+  `updates.on_schedule` so each Grid cell or Canvas element can independently
+  refresh at the server's local day boundary or an explicitly selected time.
+  Due refreshes coalesce per dashboard, respect quiet hours, update only
+  displays already showing the dashboard, and silently re-warm inactive Lineup
+  pages without selecting or advancing them. The feature is off by default.
+
 ## [0.292.3], 2026-08-12
 
 ### Fixed
--- a/app/app_factory.py
+++ b/app/app_factory.py
@@ -918,6 +918,7 @@ def create_app(
         deck_store=deck_store,
         deck_nav_store=deck_nav_store,
         page_store=page_store,
+        plugin_registry=lambda: app.config["PLUGIN_REGISTRY"],
         push_manager=lambda: app.config["PUSH_MANAGER"],
         event_log=event_log,
         timezone_provider=_resolve_timezone,
@@ -929,10 +930,13 @@ def create_app(
     app.config["SCHEDULER"] = scheduler
     from app.data_change_refresh import DataChangeRefreshCoordinator
 
+    def _refresh_data_change_pages(page_ids: set[str]) -> None:
+        scheduler.refresh_pages_for_update(page_ids, source="data_change")
+
     data_change_coordinator = DataChangeRefreshCoordinator(
         page_store=page_store,
         plugin_registry=lambda: app.config["PLUGIN_REGISTRY"],
-        refresh_pages=scheduler.refresh_pages_for_data_change,
+        refresh_pages=_refresh_data_change_pages,
     )
     app.config["DATA_CHANGE_COORDINATOR"] = data_change_coordinator
     if not testing and not is_watcher:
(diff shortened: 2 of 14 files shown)


dmellok:
Core slice reads well: the capability gate, the off-by-default posture, and the "never selects or advances" boundary are all in the right places. Five things.

**1. The retry loop has no backoff, and each quiet retry writes an event.** `refresh_pages_for_update` drops a page from `completed` unless the status is `sent`/`no_change`, so a placement blocked by quiet hours stays due and re-attempts on every tick, which is 30s by default. `PushManager._log_quiet_skip` records an event row per call, so an 8-hour quiet window is roughly 960 `quiet` rows in Events per scheduled page per night. A `failed` status is worse, since that path re-renders each time and keeps going until the local date rolls. Consider stamping the attempt and backing off, or treating `quiet` as "retry when the window ends" rather than "retry every tick".

**2. The daily-due logic is a second copy of `find_due`'s daily branch** (`app/scheduler.py:531-559`): same `replace(hour, minute)` to `astimezone(UTC)`, same `now < target`, same `first_seen` backfill guard, same `_local(last).date() == local_now.date()`. Matching the Schedule semantics is the right call, but two copies drift. Worth one helper both call.

**3. A behaviour change to the shipped data-change path rides along unmentioned.** `_page_last_refresh` is now stamped only when the status is `sent`/`no_change`; before it was stamped after any push that didn't raise. Looks like an improvement, but it changes when data-change refreshes reset the page cadence clock, so it belongs in the description.

**4. `first_seen` / `last_fired` are in-memory only.** A restart between the day boundary and the target time trips the backfill guard and drops that day's refresh. Identical to Schedules, so the consistency argument holds, but "refresh daily" reads as a stronger promise than "fires if the server happened to be up at that minute" and the docs don't say otherwise.

**5. No test for the resolver's capability gate.** `_supported()` in `scheduled_placements` is what keeps a stale schedule inert when a widget stops declaring `on_schedule`; #219 tests the route-level clear, but not this one.

Plus the CHANGELOG heading and the version bump plus `uv lock`. The repo uses versioned sections rather than `## [Unreleased]` (the PR template is stale on that), and this now conflicts with main since 0.292.2 landed.

charmmmz:
Thanks — I addressed all five items and rebased this stack onto current main.

1. Scheduled refreshes now preflight quiet hours through the public device_in_quiet_hours API, so an all-quiet dashboard remains due without calling PushManager or creating one Event row every 30 seconds. The first scheduler tick after quiet hours retries it. Actual failed/busy attempts use a five-minute retry backoff.
2. The existing Schedule path and widget-schedule path now share _daily_is_due, including target conversion, first-seen backfill, and once-per-local-day semantics.
3. I restored the shipped data-change cadence behavior: _page_last_refresh is stamped after every non-exception push result. Backoff and quiet preflight are opt-in arguments used only by scheduled refreshes.
4. The limitation is now explicit in the docs: due/satisfied state is in memory like existing Schedules, so a restart after the target can defer that placement until the next day. I left persistence for a follow-up rather than expanding this slice.
5. Added a resolver test proving a stale placement schedule becomes inert when the plugin no longer exposes the daily capability.

Also rebased onto current 0.292.3 main, created the versioned 0.293.0 changelog section, bumped pyproject.toml, and refreshed uv.lock.

Validation: 49 focused Core tests passed; the full affected stacked set is 279 passing, plus Ruff, mypy across 178 app modules, and diff checks.

PR #219, opened 2026-08-11
feat: configure scheduled widget refreshes

charmmmz:
## Summary

- add placement-level daily refresh controls to both Grid and Canvas Configure surfaces
- keep schedules off by default and store no at value for the local day-boundary behavior
- reveal the time picker progressively and persist at only after the user explicitly chooses a custom time
- let Canvas users update an existing custom time in one step, with a separate day-boundary action
- treat a blank Grid custom-time field as the day boundary so the rest of the cell form still saves
- expose suggested_at as a prefill only and source normalized schedule capabilities from the plugin accessor
- preserve schedules across partial Grid form posts with a presence marker
- clear saved schedules when a widget no longer declares the matching capability

## Dependency

Stacked on #218. Until that Core PR merges, GitHub shows its Core commits in this comparison too. After #218 lands, this PR should collapse to the Configure/UI commits only.

The Apple Reminders declaration is kept in its own widget repository PR.

Discussion: https://github.com/dmellok/tesserae/discussions/216

## User impact

A user can opt a specific widget placement into a daily refresh without changing the whole dashboard cadence. The refresh updates displays already showing the dashboard and lets Core warm inactive Lineup pages; it never selects or advances a dashboard.

## Validation

- broader affected regression set: 279 tests passed
- Ruff passed
- mypy passed across 178 app modules with the local Python 3.12 target override
- node --check static/panels/editor.js passed
- git diff --check passed
- real Canvas interaction verified: saved 07:00, changed it directly to 08:00, confirmed persisted JSON, then returned to the day boundary; no browser warnings or errors

The repository-wide frontend lint command still reports the pre-existing 13 ESLint errors and existing Prettier drift; this patch adds no new JavaScript syntax errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [0.293.0], 2026-08-12
+
+### Added
+
+- **Placement-level daily widget refreshes.** Widgets may declare
+  `updates.on_schedule` so each Grid cell or Canvas element can independently
+  refresh at the server's local day boundary or an explicitly selected time.
+  Due refreshes coalesce per dashboard, respect quiet hours, update only
+  displays already showing the dashboard, and silently re-warm inactive Lineup
+  pages without selecting or advancing them. The feature is off by default.
+
 ## [0.292.3], 2026-08-12
 
 ### Fixed
--- a/app/app_factory.py
+++ b/app/app_factory.py
@@ -918,6 +918,7 @@ def create_app(
         deck_store=deck_store,
         deck_nav_store=deck_nav_store,
         page_store=page_store,
+        plugin_registry=lambda: app.config["PLUGIN_REGISTRY"],
         push_manager=lambda: app.config["PUSH_MANAGER"],
         event_log=event_log,
         timezone_provider=_resolve_timezone,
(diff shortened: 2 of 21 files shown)


dmellok:
Two real ones here.

**1. Picking "At a specific time" and leaving the time blank rejects the whole cell save.** Verified on this branch: posting `update_schedule_timing=custom` with `update_schedule_at=""` alongside `zoom=1.4` leaves zoom at 1.0, saves no schedule, and flashes `Invalid cell: Value error, at must be HH:MM in 24-hour time`. `<input type="time">` submits an empty string whenever it's cleared or never filled, and `form.get("update_schedule_at") or ""` maps that onto a value the validator always rejects, so every other edit in that save is discarded and the user sees raw pydantic text. Falling back to `WidgetUpdateSchedule(kind="daily")` (or to `current`) when the field is blank covers it. The canvas path already guards this with `else if (timeInput.value)`, so the two surfaces disagree on the same input.

**2. A saved custom time can't be changed in one step on Canvas.** Once `at` is set, the only button reads "Use day boundary" and its handler unconditionally clears back to `{kind: "daily"}`, so editing the time and clicking throws the typed value away. Going 07:00 to 08:00 is: click (drops to boundary, `paint()` closes the details and resets the input to `suggested_at`), re-open, retype, click again. Saving `timeInput.value` when it differs from `at`, with a separate control for returning to the boundary, would fix it.

**3.** `panels_schema.catalog_entry` reads `plugin.manifest["updates"]["on_schedule"]` raw instead of `plugin.on_schedule_updates`, so the editor sees unnormalized entries while the scheduler sees filtered ones. Schema validation makes it mostly theoretical, but the property is the one source of truth.

**4.** Grid uses radios inside `<details>` and Canvas uses a toggle button for the same choice, so the two need the same fix twice. Not blocking, just noting it.

Same CHANGELOG heading and version-bump note as #218, and this conflicts with main for the same reason. Worth rebasing once #218 lands.

charmmmz:
Thanks — the three actionable issues are fixed and this PR is restacked on the updated #218 branch.

1. A blank Grid custom-time field now falls back to the local day boundary instead of rejecting the cell. The regression test posts zoom=1.4 with a blank custom time and verifies both the zoom and boundary schedule are saved.
2. Canvas now always has a Use this time action for a changed value and exposes Use day boundary as a separate action. I verified the real editor flow by saving 07:00, changing it directly to 08:00 in one step, confirming the persisted document, and then returning to the boundary.
3. The editor catalog now uses plugin.on_schedule_updates, so it sees the same normalized/filtering behavior as the scheduler.

I agree on the Grid/Canvas control consistency point; I left that as a non-blocking UI follow-up rather than broadening this review fix.

Validation: 279 affected tests passed, plus Ruff, mypy, JavaScript syntax, diff checks, and the Canvas interaction check with no browser warnings/errors.

PR #220, opened 2026-08-11
feat(gtfs): departure board for any GTFS / GTFS-RT feed

robbiet480:
## What this changes

Adds a `gtfs` widget: a departure board for any GTFS feed, with live times,
delays, cancellations, service alerts and track numbers from GTFS-RT. Two
supporting host changes come with it — a new `fill_from` cell-option
relationship, and a fix for `choices_from` dropdowns rendering empty in the
widget preview.

## Why

Wanted a lobby display showing every upcoming train at the local station.
Nothing in the bundled set reads GTFS, and the format is what practically
every transit agency publishes.

Three constraints shaped the implementation:

- **Static GTFS is too big to read inline.** A big agency's `stop_times.txt`
  runs to hundreds of MB, far past the composer's per-widget hydration
  budget. The first render kicks a background build; every render after
  reads a distilled per-stop timetable of a few KB, cached 12 hours and
  refreshed stale-while-revalidate. A failed build parks in its own file for
  10 minutes so a broken feed isn't re-downloaded on every refresh tick, and
  never clobbers a stale-but-usable table.
- **Realtime is protobuf, and the useful fields are vendor extensions.**
  Decoded directly against named field numbers rather than adding a core
  `protobuf` dependency for every install. An extension is just a field
  number the base spec doesn't claim, so the NYCT track labels
  (`gtfs-realtime-NYCT.proto`, field 1001) cost ten lines; the generated
  bindings don't ship that proto at all. Whole protobuf layer is ~170 lines
  including alerts and vehicle positions.
- **The board has to report its own uncertainty.** A stalled feed still
  serves confident-looking predictions, so the feed's header timestamp
  drives a "4 min old" note and drops the live badge past five minutes. The
  title also carries the render time: countdowns freeze at render, and an
  e-ink frame can sit for minutes.

Agencies split realtime by line group (the MTA publishes one feed per
group), so a second preset can contribute just its realtime feed, and a
board covering two stops on different groups is live on both halves.

Closes #

## Test plan

- [x] `pytest -q` — 3084 passing; 50 in `plugins/gtfs/`
- [x] `mypy app`
- [x] Manually verified against live feeds: MTA subway zip +
      `nyct/gtfs-ace` + `nyct/gtfs` + `camsys/subway-alerts`, and BART
      (`google_transit.zip`, `gtfsrt/tripupdate.aspx`, `gtfsrt/alerts.aspx`).
      Checked live times, delays in both directions, cancellations, alert
      ranking, track labels, two-stop merge (Canal St A/C/E + Franklin St 1),
      and the split-by-direction board.
- [x] Rendered through `app.quantizer` at Spectra 6 / Floyd–Steinberg rather
      than trusting the browser preview. Speckle metric 0.29 vs
      `weather_forecast`'s 0.36; route badges and headsigns stay legible at
      md and lg. Worth noting `render.dither` in a manifest is inert —
      `app/dither_regions.py` makes flat a per-cell opt-in — so the frame
      dithers regardless.
- [x] Stop finder at `/plugins/gtfs/` against the real MTA feed: search,
      inspect, and the "which direction is 0" answer.

Not covered by automated tests: everything above that needs a live feed, and
the panel render (no hardware in the loop — quantiser output was inspected,
not a physical panel).

## Checklist

- [x] Tests added for new behaviour
- [x] `ruff` + `mypy` clean
- [x] User-facing changes documented (`docs/widgets.md`, plus per-option help
      text in the manifest)
- [ ] `pyproject.toml` version bumped if this is going to be tagged — left
      alone; say the word if you want it in this PR

## Notes for review

- **`fill_from` is new host surface.** A cell option can name another option
  that owns its value; the editor fills the field and marks it read-only
  (read-only rather than disabled, so the value is still submitted and
  survives a save). Documented in `docs/widgets.md` and added to
  `schema/plugin.schema.json`. Happy to split it out if you'd rather it
  landed on its own.
- **`network:*` capability.** The feed URL is user-supplied, same case as the
  picture widgets, so no narrower declaration is possible. `_download`
  rejects any scheme but http/https and uses an opener with no
  `file://`/`ftp://` handlers, since `HTTPRedirectHandler` would otherwise
  permit an http→ftp redirect. Private/link-local destinations are *not*
  blocked: self-hosted feeds on a LAN are a first-class case here, the same
  premise `calendar_core` works on.
- **The stop finder fetches a URL from a query arg**, so an untrusted caller
  is held to the preset feeds; arbitrary URLs need a session (or an install
  that has deliberately turned the password off).
- **`plugins/gtfs/fixtures/`** back a `demo:<name>` feed URL — set the feed
  to `demo:delayed` and the board paints a canned scenario with no network,
  which is how the delay/cancellation states were designed and reviewed.

https://claude.ai/code/session_01UhJm9DoKgUA8B1FYwE2NFW

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,34 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [Unreleased]
+
+### Added
+
+- **GTFS / GTFS-RT departure board.** A new `gtfs` widget shows approaching
+  vehicles at a stop, from any agency's static GTFS zip plus optional
+  realtime feeds: live times, delays against the timetable, cancellations,
+  service alerts, and track numbers where the feed publishes them. Presets
+  cover the NYC Subway (one per realtime line group) and BART; anything else
+  takes URLs. A stop finder at `/plugins/gtfs/` searches a feed for the
+  `stop_id` a cell needs and reports which routes call there and what each
+  `direction_id` means in headsign terms, since GTFS never says. Two stops
+  can share one board — the office lobby case, a subway entrance and the
+  station round the block — and a large cell can split into two columns by
+  direction. Realtime protobuf is decoded directly, including the NYCT
+  extension, so no protobuf dependency is added.
+- **`fill_from` on a cell option.** An option can declare that another owns
+  its value; the editor fills the field from a map and locks it. The `gtfs`
+  presets use it so picking a feed fills its three URLs. See
+  [docs/widgets.md](docs/widgets.md).
+
+### Fixed
+
+- **The widget preview resolves dynamic dropdowns.** `/_test/preview` read
+  `cell_options` straight from the manifest, so any option using
+  `choices_from` rendered as an empty `<select>` — for every widget, not just
+  new ones. It now materialises choices the way the page editor does.
+
 ## [0.292.3], 2026-08-12
 
 ### Fixed
--- a/app/composer.py
+++ b/app/composer.py
@@ -2031,7 +2031,14 @@ def test_widget_preview() -> str:
     # , same schema the page editor reads. Defaults override URL-supplied
     # values only when the URL omits a field, so reloading the page with
     # an explicit blank still wins over the manifest default.
-    schema = list(plugin.manifest.get("cell_options") or [])
+    #
+    # Materialised through the page editor's helper so ``choices_from``
+    # options (the plugin resolves those at edit time) arrive with concrete
+    # choices. Reading the raw manifest here rendered every dynamic dropdown
+    # in this preview empty, for every widget that uses one.
+    from app.page_routes import _materialize_cell_options
+
+    schema = list(_materialize_cell_options([plugin]).get(plugin.id) or [])
     supplied_opts: dict[str, Any] = parsed["cell_options"]
     form_values: dict[str, Any] = {}
     for spec in schema:
--- a/app/widget_samples.py
+++ b/app/widget_samples.py
@@ -852,7 +852,54 @@ def _weather_now() -> dict[str, Any]:
     }
 
 
+def _gtfs() -> dict[str, Any]:
+    # Canal St (A/C/E), the stop the widget was developed against. Route
+    # colours are the MTA's own (data identity, not a design choice).
+    def _a(
+        minutes: int,
+        route: str,
+        headsign: str,
+        time: str,
+        live: bool,
+        delay: int = 0,
+        stop: str = "Canal St",
+    ) -> dict[str, Any]:
+        return {
+            "trip_id": f"sample-{route}-{minutes}",
+            "stop_id": "A34N",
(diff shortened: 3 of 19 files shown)


robbiet480:
Screenshots, all against live MTA feeds except 05/06 which use the bundled `demo:` fixtures. The last one is the `lg` board through `app.quantizer` at Spectra 6 / Floyd–Steinberg, which is what the panel actually shows — speckle 0.29 vs `weather_forecast's` 0.36:

01-lg-split
02-md-list
03-sm-hero
04-xs-tile
05-delays
06-cancelled
07-two-stops
08-stop-finder
09-spectra6-quantised

dmellok:
Landed on main as of v0.294.0. Closing since the commits went in as cherry-picks rather than a merge, so this won't close itself.

dmellok:
Merged, all four commits, on main as of v0.294.0. A generic GTFS reader is
broad enough to earn a place in the bundled set, so it's staying in `plugins/`
rather than going out to the catalog.

Two things I changed on the way in.

**The schema reformat.** `schema/plugin.schema.json` came in wholly
reformatted, which collided with the `updates` block from #218. I reapplied
`fill_from` and `help` by hand in the file's existing compact style instead.
Same additions, no whitespace churn.

**The opener isn't http-only.** The comment on `_OPENER` says
`build_opener(HTTPHandler, HTTPSHandler)` yields an opener without the file and
ftp handlers. It doesn't: `build_opener` installs the default handler set and
only substitutes the classes you pass, so both survive.

```
>>> op = urllib.request.build_opener(urllib.request.HTTPHandler, urllib.request.HTTPSHandler)
>>> sorted(op.handle_open)
['data', 'file', 'ftp', 'http', 'https', 'unknown']
```

The scheme check in `_download` still stops a pasted `file://` URL, so the case
you were defending against holds. The redirect case doesn't: CPython's
`HTTPRedirectHandler` permits redirects to `('http', 'https', 'ftp', '')`, and
`FTPHandler` was installed, so a feed host answering 302 could pull one.
Narrow, and it needs a hostile or compromised host, but the stated mitigation
wasn't doing the work, and it's a wider surface bundled than it would have been
as an install-it-yourself widget. Fixed in c89321c on main by popping `file`, `ftp` and
`data` out of the dispatch table, which covers the redirect path too and turns
the attempt into a `URLError` via `UnknownHandler`.

Everything else went in as written. The protobuf-by-field-number call was the
right one, the failure file that won't clobber a usable stale table is the
detail I'd have expected to be missing, and a board that reports its own
staleness is the correct instinct for a panel whose frame can sit for minutes.

One note on `fill_from` for anyone reading it later: it resolves server-side
only, so changing the controlling option and saving in one step submits the
previous fill and the field catches up on the reload after. Invisible here
because `_preset_urls` re-derives from the preset at fetch time, which is the
right shape. I wrote that into `docs/widgets.md` as the contract so the next
widget using it re-derives too.

Also worth flagging for next time: new widgets go through
https://github.com/dmellok/tesserae-widgets rather than this repo. The bundled
set in `plugins/` stays put and this one's been adopted into it, but the
catalog is the default route now. That's on me for never writing it down, so
it's in `docs/widgets.md` now.

#221 is in too.

PR #233, opened 2026-08-17
Optimize Docker layers for app-only upgrades

charmmmz:
## Summary

- install stable OS packages and Python runtime dependencies before copying frequently changing Tesserae source
- generate the Docker dependency manifest directly from `project.dependencies` on the native build platform
- let tag builds fall back to the `main` GitHub Actions cache while preserving per-ref cache writes
- keep the final editable install lightweight with `--no-deps`
- align the project/lockfile version and changelog with the repository's release convention

## Why

The current Dockerfile copies the source tree before `pip install -e /app`. A one-file app change therefore invalidates the approximately 329 MB dependency installation layer and every later layer, even when dependencies are unchanged.

Keeping system packages and runtime dependencies ahead of `pyproject.toml` and the source copies lets registries and Docker hosts reuse those blobs across ordinary app releases. The native-platform extractor avoids a second checked-in dependency list, while its generated output lets BuildKit reuse the dependency layer when only the project version changes. Tag builds also read the `main` cache so stable releases can reuse the layer produced by edge builds when GitHub still retains it.

## Impact

An isolated source-only rebuild on the current Playwright base kept the full image at approximately 2.71 GB, but reduced the uncompressed changed-layer set from approximately 359 MB to 25.8 MB (about 93%). The large dependency layer remains part of the image, but is reused rather than rebuilt and transferred for app-only changes.

## Validation

- `docker buildx build --check --progress=plain -f Dockerfile .` (no warnings)
- parsed all 17 runtime dependencies from `project.dependencies` with the extractor logic
- completed an amd64 image build from an isolated v0.309.1 source context using the reordered layers, then started an isolated container: `/healthz` returned `ok`, the root route returned the expected setup redirect, UID 1001 could write data, and Waitress started normally
- rebuilt after adding one temporary Python source file; the Playwright base, system package layer, and 329 MB dependency installation layer all hit cache
- compared image history and root filesystem layers to measure the changed-layer reduction
- `git diff --check`

The updated native-platform extractor and multi-architecture image build are also covered by this PR's Docker CI. The temporary local images and container were removed after validation. No running Tesserae deployment was replaced or restarted.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -115,6 +115,9 @@ jobs:
           # + v* tag push firing together on a release) don't race on layer
           # blob writes. Without a scope, both runs write to the same cache
           # namespace and one gets "not_found" on layer blobs the other
-          # just cleaned up. Cache reuse within a ref stays intact.
-          cache-from: type=gha,scope=${{ github.ref_name }}
+          # just cleaned up. Tags also read main's cache so stable releases can
+          # reuse dependency layers from the edge build when GHA retains them.
+          cache-from: |
+            type=gha,scope=${{ github.ref_name }}
+            type=gha,scope=main
           cache-to: type=gha,mode=max,scope=${{ github.ref_name }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [0.311.1], 2026-08-17
+
+### Changed
+
+- **App-only Docker upgrades download much smaller changed layers.** System
+  packages and Python runtime dependencies are installed before the frequently
+  changing Tesserae source, and tagged builds can fall back to the `main` build
+  cache, so unchanged large layers can be reused when the cache is available.
+
 ## [0.311.0], 2026-08-17
 
 ### Added
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,16 @@
 # for 1.60) is what the matching Python wheel goes looking for at
 # launch. A mismatch boots the container fine but errors at first
 # render with "Executable doesn't exist at /ms-playwright/...".
+
+# Generate the Docker dependency manifest from the project's source of truth.
+# Run this tiny stage on the build platform so multi-architecture builds do not
+# invoke Python under QEMU. BuildKit keys the following cross-stage COPY by the
+# generated file content, so a version-only pyproject change still reuses the
+# expensive dependency layer.
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/playwright/python:v1.60.0-noble AS dependency-manifest
+COPY pyproject.toml /tmp/pyproject.toml
+RUN python -c "import pathlib, tomllib; project = tomllib.loads(pathlib.Path('/tmp/pyproject.toml').read_text()); pathlib.Path('/tmp/requirements.txt').write_text('\n'.join(project['project']['dependencies']) + '\n')"
+
 FROM mcr.microsoft.com/playwright/python:v1.60.0-noble
 
 # Where Playwright looks for installed browsers. The base image puts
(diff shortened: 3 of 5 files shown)


dmellok (commented):
The reordering is correct and worth taking. The reasoning holds up in
particular on the point that `pyproject.toml` can't sit in the stable half,
since this project bumps its version on every release, so a dedicated manifest
is a real answer to a real constraint rather than a workaround.

One finding changes what the PR delivers, though, and it's outside the diff.

## The win is real for edge and absent for tagged releases

`.github/workflows/docker.yml` scopes the build cache per ref:

```yaml
cache-from: type=gha,scope=${{ github.ref_name }}
cache-to: type=gha,mode=max,scope=${{ github.ref_name }}
```

Consecutive pushes to `main` share scope `main`, so the dependency layer is
reused and the edge channel sees exactly the improvement you measured. But every
`v*` tag is a new `ref_name`, so a release build starts from an empty cache
scope, rebuilds every layer, and pip's fresh mtimes give the dependency layer a
new digest. Operators pulling `:latest` or `:0.311.0` still download the full
~329 MB.

The layer order is necessary for the fix; it isn't sufficient. A companion
change that keeps the blob-race fix the existing comment documents (that race
was about two runs *writing* one scope, not reading it):

```yaml
cache-from: |
  type=gha,scope=${{ github.ref_name }}
  type=gha,scope=main
cache-to: type=gha,mode=max,scope=${{ github.ref_name }}
```

Worth knowing either way: the GHA cache is 10 GB per repo with LRU eviction and
this image is ~2.7 GB at `mode=max`, so reuse across releases isn't guaranteed
even with the fallback.

## The duplicate manifest is avoidable

An extractor stage gets the same cache behaviour with one source of truth, no
drift check, and no exact-ordering brittleness:

```dockerfile
FROM --platform=$BUILDPLATFORM mcr.microsoft.com/playwright/python:v1.60.0-noble AS deps
COPY pyproject.toml /tmp/pyproject.toml
RUN python -c "import pathlib,tomllib; p=tomllib.loads(pathlib.Path('/tmp/pyproject.toml').read_text()); pathlib.Path('/tmp/requirements.txt').write_text('\n'.join(p['project']['dependencies'])+'\n')"

FROM mcr.microsoft.com/playwright/python:v1.60.0-noble
COPY --from=deps /tmp/requirements.txt /tmp/requirements.txt
RUN pip install -r /tmp/requirements.txt
```

The extractor reruns on every release because `pyproject.toml` changed, but it
costs about a second, and the `COPY --from` cache key is the content digest of
`requirements.txt`, which is unchanged when dependencies are unchanged. So the
expensive `pip install -r` still hits cache. `--platform=$BUILDPLATFORM` keeps
the extractor off QEMU during the arm64 build.

The tradeoff is genuine: a checked-in manifest is greppable and legible to
scanners, and a generated one isn't. Your call. If you'd rather keep the
explicit file, the next section is what I'd want instead.

## If the manifest stays

- **The check runs after the install it guards.** It needs
  `/app/pyproject.toml`, which is deliberately copied late, so drift costs a
  full dependency install on both architectures (arm64 under emulation) before
  failing. A short pytest asserting the two files match would catch it in the
  seconds-long `pytest` job instead. Enforcement does work today, the docker
  job ran as a `pull_request` event on this branch and passed, it's just the
  slowest place to learn about a one-line mistake.
- **The assertion is a ~380-character one-liner inside a `RUN`.** This
  Dockerfile is otherwise unusually legible; a heredoc or a file under
  `scripts/` would match its surroundings.
- `rm /tmp/requirements-docker.txt` doesn't shrink anything. The file lives in
  the earlier `COPY` layer, so the `rm` only writes a whiteout.

## Correctness, checked and clean

- All 17 entries match `project.dependencies` in content and order. Because the
  comparison is exact, a *tightened* constraint can't slip through either, which
  is what makes `--no-deps` safe here.
- `pip install -e /app --no-deps` still gets `setuptools>=68`: `--no-deps`
  suppresses runtime resolution, not PEP 517 build requirements, which come
  from the isolated build environment.
- Moving apt ahead of pip changes nothing for `gosu`, the emoji font, or the
  `chown -R pwuser:pwuser /app` that still runs last.

## Conventions

- `## [Unreleased]` isn't a heading this CHANGELOG uses. Every entry is a
  versioned heading with a date, and `pyproject.toml`'s version is bumped in
  the same commit.
- The branch is conflicting against `main` as of v0.311.0, which landed a
  CHANGELOG entry in the same region.

## Summary

Happy to take this after a rebase. The one thing I'd like in the same PR is the
`cache-from` fallback, since without it the change doesn't reach the release
channel most operators pull from. The extractor stage is a suggestion rather
than a condition.

dmellok:
Merged as 14d5632. The Dockerfile and the workflow change landed byte-identical
to your branch; the only edit was renumbering the release, since main moved to
0.312.0 while this was open, so it's 0.312.1 with the lockfile regenerated.
Cherry-picked rather than merged, which is why the SHA differs and this didn't
close itself. Authorship is intact on the commit.

Both things I asked for are in, and the extractor stage removed the duplicate
manifest and the drift check with it, which is a better shape than what I
suggested reviewing.

Two optional things I left out rather than editing your work: the "Pin the
Playwright minor" comment block now sits above the extractor stage's `FROM`
rather than the runtime one, and the pinned tag is written twice, which is a
small drift risk of the kind this PR just removed. A single
`ARG BASE_IMAGE=...` before the first `FROM` fixes both. Happy to take that as
a follow-up or do it myself.

One consequence worth naming, since it's the flip side of the change working:
reusing the dependency layer pins transitive resolution to whenever that layer
was first built. The manifest carries ranges, so previously every release
re-resolved because every release rebuilt. Now a tagged release can inherit an
edge build's resolution as long as `project.dependencies` hasn't changed
textually, which means "rebuild to pick up an upstream patch" no longer does
anything on its own. Mostly a reproducibility win, but I wanted it written down
rather than discovered.

PR #235, opened 2026-08-17
Fix prioritised multi-select reordering

charmmmz:
## Summary

- add real drag handles to selected rows in prioritised multi-select fields
- support mouse, touch, pen, and keyboard reordering
- submit the reordered DOM immediately so widget priority matches the editor
- align the compact dot handles and remove dark seams between adjacent selected rows

## Why

The help text promised drag-to-reorder, and the server already round-tripped selected values in DOM order, but the shared editor component did not provide a working reorder interaction. Priority-based widgets such as Album Art therefore could not be reordered from the UI.

## User impact

Selected media players and other prioritised multi-select values can now be reordered directly in both Dashboard editors. The resulting order is saved and used by the widget immediately.

## Validation

- `ruff check tests/test_page_routes.py`
- `pytest -q tests/test_page_routes.py` — 92 passed
- `node --check static/components.js`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [Unreleased]
+
+### Fixed
+
+- **Prioritised multi-select lists can actually be reordered.** Checked rows
+  now expose a drag handle that works with a mouse, touch, or pen, plus arrow,
+  Home, and End keys for precise accessible movement. The new DOM order is
+  submitted immediately, so priority-based widgets such as Album Art evaluate
+  Home Assistant entities in the order shown by the editor.
+
 ## [0.313.0], 2026-08-17
 
 ### Added
--- a/static/components.js
+++ b/static/components.js
@@ -193,8 +193,88 @@
       ms.dataset.msBound = "1";
       const filter = ms.querySelector("[data-ms-filter]");
       const opts = Array.from(ms.querySelectorAll(".multiselect-opt"));
+      const list = ms.querySelector("[data-ms-list]");
       const emptyEl = ms.querySelector("[data-ms-empty]");
       const countEl = ms.querySelector("[data-ms-count]");
+      const orderStatus = ms.querySelector("[data-ms-order-status]");
+      let dragged = null;
+      let dragStartOrder = [];
+      let pointerDrag = null;
+
+      const inputFor = (row) => row && row.querySelector('input[type="checkbox"]');
+      const rows = () => list ? Array.from(list.querySelectorAll(".multiselect-opt")) : [];
+      const selectedRows = () => rows().filter((row) => inputFor(row)?.checked);
+      const selectedValues = () => selectedRows().map((row) => inputFor(row).value);
+      const sameOrder = (a, b) => a.length === b.length && a.every((v, i) => v === b[i]);
+
+      function firstUnchecked() {
+        return rows().find((row) => !inputFor(row)?.checked) || emptyEl || null;
+      }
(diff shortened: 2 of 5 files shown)


dmellok:
Keyboard reordering works exactly once, then stops. I rendered the macro from
this branch into a page with its own `components.js` and `forms.css` and drove
it in headless Chromium:

```
focus:      BUTTON "Reorder Kitchen"
ArrowDown → order: [lounge*, kitchen*, study*, bedroom]   focus: BODY
ArrowDown → order: [lounge*, kitchen*, study*, bedroom]   focus: BODY   (no move)
```

`moveWithKeyboard` re-inserts every selected row through `list.insertBefore`,
including the row holding the focused handle. Moving a focused element in the
DOM blurs it, so the second arrow press lands on the page instead of the
handle. In practice a keyboard user has to re-Tab to a handle that just changed
position after every single step, which is not what the help text or this PR
promises.

The same cause hits `partitionSelection`: ticking a checkbox now moves its row,
so focus drops to `<body>` after every tick. That one is a regression against
`main`, where checking a box doesn't move anything.

I confirmed the fix in the same harness. Restoring focus after the move makes
the whole interaction behave as advertised:

```
ArrowDown  order=[lounge*, kitchen*, study*, bedroom]  focus=Reorder Kitchen
ArrowDown  order=[lounge*, study*, kitchen*, bedroom]  focus=Reorder Kitchen
Home       order=[kitchen*, lounge*, study*, bedroom]  focus=Reorder Kitchen
End        order=[lounge*, study*, kitchen*, bedroom]  focus=Reorder Kitchen
```

So: re-focus the handle after `moveWithKeyboard`, and preserve focus across
`partitionSelection`. `Element.moveBefore()` does this natively where available,
with `insertBefore` plus an explicit `focus()` as the fallback.

Neither existing test catches this because
`test_multiselect_component_wires_mouse_touch_and_keyboard_reordering` only
greps `components.js` for listener substrings. Worth a real assertion on the
post-move `document.activeElement` if there's a practical way to run it here.

The rest holds up. I traced both save paths: the dispatched `input` event is
picked up by the form listener in `static/pages/editor.js` (`_deferToBlur`
returns false for a `<div>`, so the preview refreshes and the form goes dirty),
and the Panels drawer reads `input,select,textarea` in document order in
`saveConfig`, so the new order persists either way.

Two smaller things:

- `_render_multiselect`'s new file reads use `read_text()` without
  `encoding="utf-8"`. `tests/test_ha_ingress_template.py` passes it explicitly
  for the same reason: Windows defaults to cp1252 and mangles non-ASCII.
- The CSS assertions (`radial-gradient(circle, currentColor 1.5px`,
  `"border-left" not in drag_button`, `box-shadow: none`) pin exact declarations,
  so any restyle fails the suite without a behaviour change. The `:has()` seam
  rule is worth keeping; the button's internal declarations probably aren't.

One question rather than a request: ticking a box now re-sorts the list under
the cursor, so checking several adjacent entries means each one jumps to the
bottom of the selected group before the next click. Sensible for a priority
list, but it's a visible change to a control used well beyond the prioritised
case. Was that deliberate?

dmellok:
Merged, and I applied the focus fix on the way in, so don't spend time on it.
This is on main as v0.314.0.

The three move sites now go through a small `place()` helper that uses native
`Element.moveBefore()` where the browser has it and restores focus after
`insertBefore` where it doesn't. I also added
`tests/test_multiselect_reorder_browser.py`, a live-Chromium module following
the skip pattern in `test_render_diagnostics.py`, parametrised over both paths
so the fallback is actually exercised. Confirmed it fails without the refocus.

Also switched the two `read_text()` calls to pass `encoding="utf-8"`. I left the
CSS declaration assertions alone; worth trimming next time you're in there, but
not worth churning a branch for.

The rest of the PR landed as written, drag and pointer paths included. The open
question about ticking a row re-sorting the list still stands if you have a
view on it.

PR #236, opened 2026-08-17
Add Apple Health summary server support

charmmmz:
## Summary

- advertise `health.summary` and the separate `personal_data_health` capability
- strictly validate the bounded seven-day Activity, Sleep, and Workouts snapshot before latest-only storage
- emit source-wide semantic refresh events while ignoring envelope-only renewal
- keep personal-data snapshots out of backups and broaden the operator-facing scope description beyond Reminders

## Privacy and lifecycle

- accepts only the contract fields; raw samples, routes, heart rate, HealthKit identifiers, source/device identity, and free-form metadata are rejected as unexpected fields
- enforces the existing 48-hour maximum TTL and a 256 KiB request limit
- retains one expiring snapshot per publisher and source, separate from History
- returns path/rule-only validation errors without echoing submitted Health values

## Contract and discussion

- Implements the merged contract from [Companion PR #7](https://github.com/charmmmz/tesserae-companion-ios/pull/7).
- Follows the server checklist accepted in [Discussion #176](https://github.com/dmellok/tesserae/discussions/176#discussioncomment-18044602).
- The Apple Health widget remains a separate community-catalog repository change.

## Validation

- `python -m pytest -q tests/companion/test_companion_personal_data.py tests/companion/test_companion_api.py tests/companion/test_contract_fixtures.py tests/test_data_change_refresh.py tests/test_backup.py` — 135 passed
- `mypy --python-version 3.12 app/health_summary.py app/companion_api.py app/data_change_refresh.py app/state/personal_data_store.py app/backup.py` — passed
- `ruff check` and `ruff format --check` on the changed Python files — passed
- `git diff --check` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [Unreleased]
+
+### Added
+
+- **Apple Health seven-day summaries can now be published through Companion.**
+  Servers advertise the opt-in `health.summary` source, strictly validate its
+  bounded Activity, Sleep, and Workout sections, retain only the latest
+  publisher-isolated snapshot until expiry, and emit source-wide semantic
+  refresh events without putting raw Health values in logs, API errors, or
+  backups.
+
 ## [0.313.0], 2026-08-17
 
 ### Added
--- a/app/app_factory.py
+++ b/app/app_factory.py
@@ -1035,8 +1035,9 @@ def create_app(
     from app.state.personal_data_store import PersonalDataSnapshotStore
 
     app.config["COMPANION_PAIRING_STORE"] = PairingStore()
-    # Personal-data bridge (#176): latest-only Apple-Reminders snapshot the
-    # Companion publishes, kept off any render/History path.
+    # Personal-data bridge (#176): latest-only, expiring Reminders and Health
+    # snapshots published by Companion. The store is separate from History
+    # and explicitly excluded from backups.
     app.config["PERSONAL_DATA_STORE"] = PersonalDataSnapshotStore(
         data_root / "core" / "companion_personal_data.json"
     )
--- a/app/backup.py
+++ b/app/backup.py
@@ -45,12 +45,15 @@ LABEL_PRE_UPDATE = "pre-update"
 #   composition PNGs and per-renderer ``.bin`` artifacts. These regenerate
 #   the moment any dashboard is pushed again, and the push code already
 #   handles the "PNG evicted from disk" case for old history entries.
+# - ``core/companion_personal_data.json`` contains latest-only personal values
+#   explicitly excluded from backups by the Companion privacy contract.
 #
 # Restoring a backup that excluded a subpath does NOT wipe the user's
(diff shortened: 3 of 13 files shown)


dmellok:
One fix needed, then this is good to go.

**A crafted `data.time_zone` raises an unhandled exception instead of a 400.**
`_TIME_ZONE_RE` allows `..` as a path component, and `ZoneInfo` raises
`ValueError` (not `ZoneInfoNotFoundError`) for a key that escapes TZPATH, so it
propagates out of the handler. Against this branch:

```
PUT /api/app/v1/personal-data/health.summary   data.time_zone = "../../etc/passwd"
ValueError: ZoneInfo keys must refer to subdirectories of TZPATH, got: ../../etc/passwd
```

`".."` alone does the same. Authenticated-only and nothing leaks, but it's a 500
with a traceback where the contract says 400. The tidiest fix is to move the
`time_zone != active_timezone` comparison above the `ZoneInfo()` call, so the
constructor only ever sees the instance's own zone. Catching
`(ZoneInfoNotFoundError, ValueError)` works too, but the reorder means untrusted
input never reaches `ZoneInfo` at all.

Everything else checks out. I ran both vendored fixtures
(`personal-data-health-summary.json` and `-partial.json`) through
`validate_health_summary` directly and both pass, so the validator and the
contract agree on the payloads the app actually sends. The strict-fields
approach, the privacy-safe error messages, and the source-wide diff that ignores
envelope-only renewal all match what we agreed on the 16th, and the backup
exclusion is the right call.

Two non-blocking notes:

**The validator enforces rules the contract doesn't state.** The `move_mode`
cross-field rules (`active_energy` requires null move minutes, `move_time`
requires null active energy, null `move_mode` requires every ring value null)
and "`segments_truncated: true` requires empty `segments`" appear nowhere in
`HealthActivityDay` / `HealthWorkout` or their descriptions. Your fixtures comply,
so nothing breaks today, but a contract-valid payload from a later Companion
build can still 400. Worth adding to the yaml descriptions so both sides encode
the same rules. Related: every failure returns `invalid_snapshot`, so the app
can't tell "your server's timezone differs from this phone's" (which needs a
user prompt) from a malformed payload (which needs a bug fix). A distinguishable
reason for the timezone case would help the app behave sensibly.

**Zero-length intervals.** `_interval` requires `end > start`, while the contract
allows `duration_seconds: 0` with no constraint relating the two. A workout
recorded with identical start and end is contract-valid and server-rejected.
Probably fine to leave, just flagging the asymmetry.

On merge mechanics: the CHANGELOG entry lands under a new `## [Unreleased]`
heading. This file goes straight to versioned headings, and #235 and #238 each
add the same one, so all three will conflict. I'll fold it into the release
heading when this goes in.

dmellok:
Merged, and I applied the timezone fix on the way in, so don't spend time on it.
This is on main as v0.314.0.

I moved the `time_zone != active_timezone` comparison above the `ZoneInfo()`
call so only the instance's own key is ever constructed, and widened the except
to `(ZoneInfoNotFoundError, ValueError)` as a backstop. Added a parametrised
test over `..`, `../../etc/passwd`, and an unknown zone asserting a 400.

Everything else landed as written. The two non-blocking notes stay open as
follow-ups whenever suits: documenting the `move_mode` and `segments_truncated`
cross-field rules in the contract yaml, and giving the timezone-mismatch case a
reason the app can distinguish from a malformed payload.

The Apple Health catalog widget is unblocked against a live source now.

PR #238, opened 2026-08-17
Add Companion device setup pairing

charmmmz:
## Summary

Let an authenticated Companion app mint a short-lived firmware registration
code for a physically nearby display during Bluetooth onboarding.

## What changed

- advertise the additive `device_setup` Companion capability
- add `POST /api/app/v1/device-pairings`, protected by
  `device_setup:write`
- issue a normal firmware pairing code from the existing `PAIRING_STORE`; the
  display still redeems it through `/api/v1/device/register`
- include device setup permission on new Companion pairings without silently
  granting it to existing stored pairings
- let an operator grant or revoke the permission per paired Companion in
  Settings
- update the vendored Companion OpenAPI contract to 0.14.0
- cover authorization, registration, and single-use replay rejection

## Why

The Companion should not receive or manufacture a display access token. This
keeps the existing firmware registration boundary intact: the app requests an
ephemeral code, transfers it locally over authenticated BLE, and the display
redeems it directly with Tesserae.

## Security properties

- requires an authenticated Companion bearer with a dedicated scope
- firmware pairing codes remain short-lived and single-use
- Companion and firmware pairing stores remain separate
- the app never receives the resulting firmware credential
- existing Companion sessions do not gain the new permission on server upgrade

## Validation

- `pytest tests/companion/test_companion_api.py tests/companion/test_contract_fixtures.py tests/test_page_routes.py -q` — 187 passed
- `ruff check` on changed Python files
- `ruff format --check` on changed Python files
- `mypy app` — 189 source files passed
- `git diff --check`

## Related firmware

- dmellok/tesserae-device-firmware#11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions are
 [SemVer](https://semver.org/) (pre-1.0, so minors can carry breaking changes).
 
+## [Unreleased]
+
+### Added
+
+- **Companion can securely register a physically nearby display during
+  Bluetooth setup.** Authenticated app sessions with `device_setup:write` may
+  mint a short-lived, single-use firmware registration code through
+  `POST /api/app/v1/device-pairings`. Firmware still redeems that code through
+  the existing device registration endpoint, so the app never receives a
+  display token. Existing Companion sessions require an explicit operator
+  grant; new pairings include the setup permission.
+
 ## [0.316.0], 2026-08-18
 
 ### Added
@@ -80,7 +92,6 @@ All notable changes to Tesserae are recorded here. Format loosely follows
   Home Assistant entities in the order shown by the editor. Reordering keeps
   the handle focused across the move, so arrow keys step an entry more than
   once and ticking a row doesn't drop keyboard focus to the page.
-
 ## [0.313.0], 2026-08-17
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ All notable changes to Tesserae are recorded here. Format loosely follows
   mint a short-lived, single-use firmware registration code through
   `POST /api/app/v1/device-pairings`. Firmware still redeems that code through
   the existing device registration endpoint, so the app never receives a
-  display token. Existing Companion sessions require an explicit operator
-  grant; new pairings include the setup permission.
+  display token. Each Companion session keeps at most one unredeemed code;
+  requesting another invalidates the previous one. Existing Companion sessions
+  require an explicit operator grant; new pairings include the setup permission.
 
 ## [0.316.0], 2026-08-18
 
--- a/app/companion_api.py
+++ b/app/companion_api.py
@@ -566,7 +566,10 @@ def create_device_pairing() -> Any:
     so the app never receives a firmware access token and this route cannot
     directly create a device.
     """
-    record = _firmware_pairings().issue(note=f"Companion: {g.companion.name}"[:64])
+    record = _firmware_pairings().issue(
+        note=f"Companion: {g.companion.name}"[:64],
+        owner_key=f"companion:{g.companion.token_id}",
+    )
     expires_at = datetime.fromtimestamp(record.expires_at, UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
     return jsonify({"code": record.code, "expires_at": expires_at}), 201
 
--- a/app/companion_api.py
+++ b/app/companion_api.py
@@ -146,6 +146,7 @@ _REMINDER_PRIORITIES = frozenset(("none", "low", "medium", "high"))
 # assuming the full set, so unshipped surfaces degrade cleanly.
 FEATURES = (
     "devices",
+    "device_setup",
     "dashboards",
     "dashboard_push",
     "image_push",
@@ -208,6 +209,10 @@ def _tokens() -> CompanionTokenStore:
 
 def _companion_pairings() -> PairingStore:
     return current_app.config["COMPANION_PAIRING_STORE"]  # type: ignore[no-any-return]
+
+
+def _firmware_pairings() -> PairingStore:
+    return current_app.config["PAIRING_STORE"]  # type: ignore[no-any-return]
 
 
 def _devices() -> DeviceRegistry:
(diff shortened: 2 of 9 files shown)


dmellok:
Reviewing the draft, since the shape is worth confirming before you finish it.

The boundary is right. Minting an ephemeral code and letting the display redeem
it directly is the version of this that doesn't hand the app a firmware
credential, and keeping the two pairing stores separate is what makes that hold.
Four things.

**1. `writeOnly: true` on `FirmwareDevicePairing.code` contradicts the response.**
The schema is only ever a 201 response body and `code` is in `required`. OpenAPI
3.0 says a `writeOnly` property MAY be sent in a request and SHOULD NOT be sent
in a response, so the contract currently declares that the one field the client
needs isn't returned. `jsonschema` ignores the keyword so `_validate` passes and
CI stays green, but a strict generator can legitimately drop `code` from the
generated response model. If the intent was "don't log this", put that in the
description and drop the keyword.

**2. Nothing caps outstanding codes.** `PairingStore.issue()` has no limit and
the new route adds none, so a paired app (or a buggy one that mints on every
screen appearance) can grow the live set for the full ten-minute TTL. Two
consequences: the Settings pending-code list fills with `Companion: <name>`
entries, and each extra live code raises the per-guess hit rate against
`/api/v1/device/register`. The register limiter (10 failed attempts per IP per
60s) keeps that from being alarming, so this is about hygiene rather than a
break. The cheap fix also matches the admin UX: hold at most one live
Companion-issued code per token and revoke the previous one on re-issue, so the
"you already have a code waiting" affordance behaves the same on both paths.

**3. Both security properties in the description are untested.**
`test_device_pairing_requires_dedicated_scope` mutates `record.scopes` in place,
which bypasses `set_optional_scope` and the persistence path, so the operator
revoke route the description promises isn't actually exercised.
`test_companion_lineups.py` has the pattern for going through the store. And
nothing asserts that a pairing persisted before the upgrade doesn't gain
`device_setup:write` on load. That's the property most worth pinning, because
it's the one a later refactor could silently break.

**4. `device_setup:write` sits in both `COMPANION_SCOPES` and `OPTIONAL_SCOPES`.**
The comment explains the intent and I think it's fine, but note the UI
consequence: `abilities` filters out everything in `OPTIONAL_SCOPES`, so this
permission never appears in the plain "what this app can do" list. It shows only
as a toggle that happens to already be on. That reads correctly to me. Confirm
it's what you intended.

Nits:

- `"Set up nearby displays"` is good phrasing; the plural slightly oversells it,
  since each call covers one display. "Set up a nearby display" tracks the
  endpoint.
- The CHANGELOG entry lands under a new `## [Unreleased]` heading. This file goes
  straight to versioned headings, and #235 and #236 each add the same one, so all
  three will conflict. I'll fold it in at merge.

dmellok:
Context for this branch: #235 and #236 are merged and on main as v0.314.0. I
applied the two fixes from those reviews on the way in, so neither needs
anything from you.

That means this branch's CHANGELOG entry now conflicts with the 0.314.0 heading
and needs a rebase. Worth knowing before you do it: a few more fixes for other
issues are landing today, so the version number will keep drifting and any
heading you pick here will be stale by the time this merges. Easiest path is to
rebase and leave the entry under `## [Unreleased]`, and I'll fold it into
whatever the release heading is at merge time. No need to chase the number.

The four review points above are still open, and none of them are urgent. The
`writeOnly` keyword on `FirmwareDevicePairing.code` is the one worth doing
before this comes out of draft, since it is the item that can affect a
generated client rather than just the server.

charmmmz:
Thanks — I rebased this branch onto current `main` and addressed all four points in `91165aa6`:

1. Removed `writeOnly: true` from the response-only firmware pairing code and documented that the returned value must not be logged.
2. Added an opaque per-token owner key to pending firmware pairings, so reissuing from the same Companion token invalidates its previous live code while admin-issued codes retain their existing behavior.
3. Updated the scope-revocation test to use `set_optional_scope()` and added a persisted pre-upgrade pairing test that pins the no-silent-grant behavior. I also added route/store coverage for replacement codes and a contract assertion guarding the response schema.
4. Confirmed the permission UI behavior is intentional: new pairings receive `device_setup:write`, existing pairings do not gain it on load, and operators can grant or revoke it; showing it only as the optional toggle avoids duplicating it in the abilities list.

I also changed the label to “Set up a nearby display” and kept the changelog entry under `Unreleased` after the rebase.

Local validation: full Ruff check and format check, `mypy app` (190 source files), and 119 focused pairing/Companion/contract tests.

dmellok:
Merged and it's on edge now as v0.317.0 (`ghcr.io/dmellok/tesserae:edge`).

Before I tag a stable release I'd like the device setup pairing exercised against the real app, since it went in right at the end of the batch and hasn't had any edge soak. If you can run a pairing pass when you get a chance and let me know how it goes, I'll cut the release off the current head.

Also on edge in the same batch, in case it's useful to test alongside: the Apple Health summary endpoint, native Gallery browse/create/upload with Send hand-off, offline album authoring, and the per-display next-visible-updates projection.

charmmmz:
Done — I ran a full pairing pass against `edge` (v0.317.0) with a real
reTerminal E1004 and Companion v0.7.0. Results below.

**Environment**
- Server: `ghcr.io/dmellok/tesserae:edge` (v0.317.0)
- Firmware: [firmware version / commit, e.g. v1.13.0 from the merged BLE setup branch]
- Companion: v0.7.0 (Nearby BLE setup UI)
- Device: reTerminal E1004 + iPhone [model / iOS version]

**What was exercised**
1. Companion BLE discovery of the nearby display — ✅ 
2. Pairing-code minting through the new `device-pairings` endpoint and
   transfer over BLE (QR / 6-digit) — ✅ 
3. Wi-Fi configuration and display registration with the server — ✅ 
4. Reconnect with the existing server registration preserved — ✅
5. Diagnostics (device identity, live IP/RSSI, maintenance events) — ✅
6. Restart — ✅
7. Clear Wi-Fi (kept distinct from factory reset) — ✅

No failures or anomalies observed. You can cut the stable release off the
current head whenever you're ready.

867E9445-9567-4A2D-8E17-392D32E084E8_1_102_o

F8947F47-7A8C-4257-81E9-E6BE2CE2EA7F

5990E89B-E8FB-4C59-A971-E74CD2DBAB7E

BCD22D24-97AB-4988-B550-EA2B0810E3E7_4_5005_c

04E30018-344F-4DBA-B4A7-8366CA713AFA

7FBD801A-64DA-4FD9-8238-EDFF3EBE29E1

dmellok:
Nice! I'll test with my e1001-3 tomorrow morning then cut the release!

Issue #255, opened 2026-08-22
[Bug]: Artwork suddenly does not render anymore (failed image)


stefbontinckx:
### Tesserae version

v0.335.0

### How is Tesserae running?

Docker compose

### Operating system

Debian 12

### Panel hardware

E1002

### Steps to reproduce

1. Use the Home Assistant - Album Art widget
2. Image renders in Tesserae dashboard canvas
3. Image does not render in sending it to the device 

Image is not included in the History overview, and not on device itself. 

E1002 was updated automatically to v1.17.0 this morning. 
Issues was not present before today (but I have updated both Tesserae server and device firmware yesterday as well) 

<img width="835" height="521" alt="Image" src="https://github.com/user-attachments/assets/9b741d82-0209-42d6-aa80-7384953b9c51" />

<img width="961" height="546" alt="Image" src="https://github.com/user-attachments/assets/e6d3659b-c16d-4d81-b68c-ac491acb108d" />

### Expected behaviour

Image should be rendering and sent to the device. 
Now it renders a 'image failed' icon, while the widget and dashboard should have an image. 

### Actual behaviour

Image fails to be shown. 

### Server logs

```shell

```

### Before submitting

- [x] I searched [existing issues](https://github.com/dmellok/tesserae/issues?q=is%3Aissue) and didn't find a duplicate.
- [x] If I could, I tried reproducing on the latest minor release.

dmellok:
The two screenshots and the "renders in canvas but not on send" split were the useful part, that narrows it a lot.

The canvas preview loads the album art in your browser, on your network. The frame that goes to the panel is rendered by headless Chromium inside the Tesserae container. Both ask Home Assistant for the same `entity_picture` URL, so the image is failing specifically for the container.

Some things I could rule out from the code:

- The album art widget points straight at your Home Assistant host rather than at a Tesserae route, so a recent change to how plugin routes are authenticated is not involved.
- The renderer already waits for every image, including inside each widget's shadow root, so this is not a race with the screenshot.
- Your widget is showing track metadata, which means the container reaches Home Assistant's API fine. So it is the image fetch specifically, not the connection generally.

What I could not tell you is *why* it failed, because nothing recorded it. The image wait deliberately treats an errored image as settled, so that a dead upstream cannot hold a render hostage, but that meant a frame full of broken images completed with the server logging a success. That is why your Server logs box is empty; there was genuinely nothing to paste.

[...]

PR #256, opened 2026-08-23
docs(widgets): document RENDER_SAFE_ENDPOINTS for plugins serving their own media

### LINKED ISSUE #253: Document RENDER_SAFE_ENDPOINTS for plugins that serve their own media
v0.329.0 gated plugin blueprint routes behind the password gate (they previously answered unauthenticated requests from loopback). The loopback exemption now covers only the static asset handler, plus any endpoint a plugin explicitly declares:

```python
# plugins/<id>/server.py
RENDER_SAFE_ENDPOINTS: tuple[str, ...] = (
    "picture_gallery_admin.serve_image",
    "picture_gallery_admin.serve_thumbnail",
)
```

The plugin loader collects these into `app.config["RENDER_SAFE_ENDPOINTS"]` and `app/auth.py` consults it.

## Why it needs documenting

A widget whose markup points at its own plugin's route needs that route served during a render, which happens over loopback with no session. `picture_gallery` is the in-tree example: its widget data hands the panel an `` pointing at `serve_image`, and without the declaration every gallery photo on every panel comes back broken.

A third-party widget author hitting this sees broken images on the panel while the browser preview looks fine, with nothing to search for. The convention is currently discoverable only by reading `app/auth.py` or the gallery plugin.

## Scope

Add it to the plugin authoring docs next to the `blueprint()` contract, covering:

- what the declaration is for (media a rendered widget references)
- that it is opt-in and should stay read-only: anything that writes, deletes, or reaches the network does not belong in the list
- that reviewing the list is part of reviewing the plugin

vaibhav8a:
Closes #253.

Adds a `Serving your own media, RENDER_SAFE_ENDPOINTS` section to `docs/widgets.md`, directly after the `blueprint()` contract as the issue asks. Covers the three points requested: what the declaration is for, that it is opt-in and must stay read-only, and that reviewing the list is part of reviewing a plugin.

## One addition beyond the issue

The issue frames the declaration as the thing standing between a widget and broken images. Reading `app/auth.py`, that is no longer the whole picture — `_path_is_loopback_only` grants a loopback `GET`/`HEAD` of a plugin's **own sub-route** without any declaration (rule 3, added in #255).

So documenting only "declare it or your images break" would have been wrong for the common case. Instead the section says the declaration may not be needed, and names what the sub-route rule does *not* cover:

- the plugin **index** (`/plugins/<id>/`), which stays gated even on a read (#221)
- routes outside the `/plugins/<id>/…` prefix
- being explicit, so the requirement survives a later tightening of the gate

with the recommendation to declare media routes anyway.

Each of those claims is pinned by a test already in the tree, which is how I checked them rather than reasoning from the code:

| claim | test |
|---|---|
| GET of an own sub-route passes with `frozenset()` | `test_a_plugin_can_serve_its_own_media_to_the_renderer` |
| plugin index refused even on a read | `test_a_plugin_index_is_still_refused_even_on_a_read` |
| writes refused regardless | `test_a_write_to_a_plugin_route_is_still_refused` |

The read-only warning uses the concrete examples from `auth.py`'s own docstring (`shutil.rmtree` on a gallery folder, calendar feed writes, CalDAV discovery), so the docs and the rationale in the code say the same thing.

Documentation only — no code touched.

### DIFF
--- a/docs/widgets.md
+++ b/docs/widgets.md
@@ -315,6 +315,46 @@ keeps Flask from colliding with the host's `templates/index.html`).
 Extend `_base.html` for visual consistency with the rest of the
 admin UI; the `card_head` macro + `card` class wrap content.
 
+### Serving your own media, `RENDER_SAFE_ENDPOINTS`
+
+Blueprint routes sit behind the password gate. A render happens over
+loopback with no session, so a widget whose markup points at its own
+plugin route — an `<img src>` handed to the panel, say — needs that
+route reachable during the render, or every panel shows a broken image
+while the browser preview looks fine.
+
+A plugin declares such routes by name:
+
+```python
+# plugins/<id>/server.py
+RENDER_SAFE_ENDPOINTS: tuple[str, ...] = (
+    "picture_gallery_admin.serve_image",
+    "picture_gallery_admin.serve_thumbnail",
+)
+```
+
+The loader collects these into `app.config["RENDER_SAFE_ENDPOINTS"]` and
+`app/auth.py` consults them. `picture_gallery` is the in-tree example.
+
+**Read-only, always.** Anything that writes, deletes, or reaches the
+network does not belong in this list. The routes found when this was
+tightened were mutations — `shutil.rmtree` on a gallery folder, calendar
+feed writes, CalDAV discovery against a caller-supplied URL — and those
+stay gated. Reviewing the list is part of reviewing a plugin.
+
+**You may not need it.** A loopback `GET`/`HEAD` of a plugin's own
+sub-route already passes without a declaration. What that rule does not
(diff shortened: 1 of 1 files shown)


dmellok:
Merged, thanks. The sub-route addition was the right call. Documenting this as "declare it or your images break" would have been wrong for the common case, and pinning each claim to an existing test rather than reading the code is the way to check it.

One follow-up in ca5801c: the section documented the read-only constraint on the declared list, but the same constraint applies to routes you don't declare, since the loopback `GET` reaches any sub-route without a declaration. Every in-tree plugin keeps mutations on `POST`, so nothing is broken today, but a catalog author had no way to know it mattered. Out in v0.352.1.

PR #264, opened 2026-08-27
feat(widgets): calendar_month translated to french

theiereman:
## What this changes

Translates the calendar_month widget to french

image

## Why

Make the widget accessible to french speaking users

## Test plan

- [ ] `pytest -q` -> still unable to run the tests under 1.5h without crashing
- [x] `mypy app`
- [x] Manually verified: every part of the widget is correctly translated to french

## Checklist

- [x] Tests added for new behaviour (or note why none made sense)
- [x] `ruff` + `mypy` clean
- [ ] User-facing changes documented (README, `docs/`, or both)
- [x] CHANGELOG entry added under `## [Unreleased]` if user-facing
- [x] `pyproject.toml` version bumped if this is going to be tagged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 - **The Calendar (week) widget is translated to French.** Header, event counts, and the
   weekday / month labels follow the render locale; non-English names come from `Intl`
   rather than a hand-rolled table.
+- **The Calendar (month) widget is translated to French.** Month title, weekday header,
+  and the week-start note follow the render locale, on the same `Intl` + `strings/` path
+  as the week widget. The weekday header row now upper-cases by default like the other
+  calendar labels (sentence-case styles still get sentence case via `--label-transform`).
 
 ## [0.367.3], 2026-08-27
 
--- a/plugins/calendar_month/client.js
+++ b/plugins/calendar_month/client.js
@@ -26,19 +26,44 @@ function escapeHtml(s) {
   }[c]));
 }
 
-const DOW = ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"];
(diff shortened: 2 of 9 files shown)


dmellok:
Merged, along with #262 and #263. Thanks for all three; the preview picker especially will make future translations much easier to check.

A few tweaks I made while merging, so you know what changed on your branches:

- All three needed `uv.lock` regenerated (CI checks it against the pyproject version), and fresh version bumps since the PRs raced each other for 0.368.0. They landed as 0.368.0–0.370.0, all on edge now.
- #263: the CHANGELOG had picked up a full-file reformat (looks like a Prettier pass, `*italics*` rewritten to `_italics_`), which I reverted to just your entry.
- #263/#264: I took the `.toUpperCase()` out of `styleLabel` and left display casing to CSS via `--label-transform`, with that rule added to `.mc-dow`. Same look by default, but sentence-case styles like Editorial keep their sentence case.
- #262: unknown `?locale=` tags are now dropped instead of forwarded, since an invalid tag would make `Intl.DateTimeFormat` throw client-side.

One limitation I left as-is: the generic abbreviation fallbacks collide in French (juin/juillet both shorten to "jui", mars/mai to "ma"). Fine for now; per-locale short/minimal maps would make a nice follow-up if you're interested.

On the test suite taking 1.5h: with pytest-xdist it runs in about 5 minutes here (`pytest -q -n auto`), worth a try.

theiereman:
Feedback greatly appreciated, thank you! I tried the test suite with `-n auto` and can confirm it make it **A LOT** faster.

I'm aware of the collisions and think that it's not a critical issue as people generally know the current month without having to rely heavily on a calendar. But still, worth getting fixed in later releases with a custom mapping as suggested.

dmellok:
You're right I was overthinking the collisions. Looking forward to future contributions :)

PR #267, opened 2026-08-29
fix(calendar_month): spread multi-day timed events across every day they cover

### LINKED ISSUE #213: calendar_month buckets multi-day timed events on their start day only
`calendar_week` learned to spread a multi-day *timed* event across every day it covers (#210), using the shared `timed_event_date_keys` helper in `app/calendar_time.py`. `calendar_month` still buckets those on their start day only:

```python
# plugins/calendar_month/server.py:69
day_keys = [event_local_date_key(ev, zone)]
```

So a three-day trip that starts Friday 16:00 and ends Monday 10:00 shows on Friday's cell and nowhere else, and the same event in the week view now correctly spans four columns. All-day events were already handled correctly in both.

The helper exists and the change is two lines. It's filed separately rather than folded into #210 because it is a visible behaviour change for anyone already using the month view: a multi-day trip would start occupying three or four cells where it used to occupy one, and a busy month grid has much less room per cell than a week column. That deserves its own look at how the month cell renders a spanning event (and possibly a continuation marker rather than a repeated title, which is what the week view ended up needing).

vaibhav8a:
Fixes #213.

## The change

`calendar_month` bucketed timed events on their start day only:

```python
day_keys = [event_local_date_key(ev, zone)]
```

so a trip starting Friday 16:00 and ending Monday 10:00 showed in Friday's cell and nowhere else, while the same event correctly spanned four columns in the week view. This switches it to the shared `timed_event_date_keys` helper `calendar_week` already uses, which also brings the exclusive-end convention with it — a block ending exactly at local midnight stops on the previous day rather than spilling.

## On the rendering question

You flagged that this is a visible behaviour change and that a month cell has far less room than a week column, so a spanning event probably wants a continuation marker rather than a repeated title. Agreed, so that is included rather than left for later:

- The server now attaches `continued` to each per-day copy — true on every day of the span except the one the event starts on. It is computed against the event's **own** start date, not the first visible key, so a span entering from the previous month is correctly marked as a continuation rather than looking like it starts at the grid edge.
- The `text` cell renders a pass-through with a leading `↳` and `opacity: 0.72`, keeping the untouched title in the `title` tooltip so the cell stays identifiable. Opacity rather than a lighter colour, so it degrades the same way as the rest of the grid on 1-bit and greyscale panels.
- The default **bars** mode is unchanged: a colour strip already reads as presence rather than as a fresh occurrence, so it needed nothing.

All-day events also carry the flag, since they were already spread and had the same repeated-title problem in the text cell — it would have been odd for the two kinds to render differently on a pass-through day.

## Tests

Two added to `tests/test_calendar_widget_timezone.py`:

- `..._spreads_multi_day_timed_events_across_covered_dates` — the reported case. Fails on `main` with `assert [] == ['Trip']` on the second day, passes with the fix, and pins the `continued` flags.
- `..._timed_event_ending_at_midnight_does_not_spill` — guards the exclusive-end edge the helper brings in.

Verified: `pytest -k calendar` 49 passed (one unrelated collection error in `test_plugin_route_auth.py` from a missing `paho` MQTT dep locally), `node plugins/calendar_month/tests/clamp_check.mjs` passes, and `ruff check .` + `ruff format --check .` are clean repo-wide on ruff 0.15.22.

### DIFF
--- a/plugins/calendar_month/client.js
+++ b/plugins/calendar_month/client.js
@@ -176,7 +176,14 @@ export default function render(shadow, ctx) {
           const colour = ev.colour || "var(--accent-4)";
           const location = showLocations && ev.location ? ev.location : "";
           const text = location ? `${ev.summary || ""} · ${location}` : (ev.summary || "");
-          return `<span class="mc-text" style="border-left-color:${colour}" title="${escapeHtml(text)}">${escapeHtml(text)}</span>`;
+          // A multi-day event occupies every cell it runs through. A month
+          // grid has far less room per cell than a week column, so repeating
+          // the full title in each is mostly noise: mark the pass-through
+          // days with a leading arrow. The untouched title stays in the
+          // tooltip, so the cell is still identifiable.
+          const label = ev.continued ? `↳ ${text}` : text;
+          const cls = ev.continued ? "mc-text is-continued" : "mc-text";
+          return `<span class="${cls}" style="border-left-color:${colour}" title="${escapeHtml(text)}">${escapeHtml(label)}</span>`;
         }).join("")}
         ${remainder > 0 ? `<span class="mc-more">+${remainder}</span>` : ""}`;
     } else {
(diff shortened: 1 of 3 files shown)


dmellok:
Merged, on edge as v0.375.0. The continuation marker went in as-is; the opacity choice reads well in the browser renderer, and if the ↳ glyph turns out to render poorly on any of the 1-bit panels I'll swap it for something plainer. The month-entry edge case in the tests was a good catch. Thanks.

Issue #214, opened 2026-08-10
calendar_schedule: slider bounds and an unrun JS self-check


dmellok:
Two things from the review of dmellok/tesserae-widget-calendar-schedule#5, neither blocking at the time.

### The size sliders lost their guard rails

`event_title_scale`, `event_time_scale` and `event_location_scale` went from `0.7–1.5` (and `0.6–1.3` for location) to `0.01–10.0`. At 0.01 the text is invisible; at 10 a single word fills the cell. The clamps still work, so nothing breaks, but the usable range is now a sliver of the travel and the ends are states no one wants.

Worth revisiting whether the widened range solved a real case (someone wanting 2x on a large panel, say) that a narrower `0.5–3.0` wouldn't. The same applies to `day_row_padding_em`, now `0.0–3.0`.

Since v0.287.0 every slider also has a number box next to it, so a wide range is less necessary: a user who genuinely wants 4x can type it even if the slider stops earlier.

### `clamp_check.mjs` isn't run by anything

The widget ships `tests/clamp_check.mjs`, a plain-node assertion script whose header says `Run: node tests/clamp_check.mjs`. There's no CI workflow and no `package.json` in that repo, so it only ever runs if someone remembers. It'll drift from `client.js` and nobody will notice.

Options: a two-line GitHub Actions workflow in the widget repo, or a `package.json` with a `test` script, or fold the assertions into the Python smoke test via a subprocess call so the existing pytest run covers them.

Both are in the marketplace widget repo, not this one; filed here because that's where the backlog lives.

PR #268, opened 2026-08-29
test: run the widget clamp_check self-checks under pytest

vaibhav8a:
Refs #214 — the second half of it, applied to this repo rather than the marketplace one.

## What I found

#214 says `clamp_check.mjs` "isn't run by anything" in the calendar-schedule widget repo. The same is true here, three times over:

```
plugins/calendar_month/tests/clamp_check.mjs
plugins/calendar_day/tests/clamp_check.mjs
plugins/calendar_week/tests/clamp_check.mjs
```

Nothing references any of them — not `.github/workflows/`, not pytest, not a `package.json` script. They pass today; they have simply been free to drift from `client.js` with nothing to notice.

`calendar_day`'s own header records how the hole opened:

> Mirrors the cases previously covered by `test_smoke.py`'s `test_scale_sliders_clamp_to_bounds` before the clamp moved from `server.py` to `client.js`.

So these assertions used to run in CI, and stopped when the clamp moved to the client. That is a coverage regression with a paper trail rather than a gap that was never filled.

## The approach

The third option from your list — fold them into pytest via a subprocess call. The scripts stay next to the code they assert on, and the existing `pytest` job executes them; no new workflow, no `package.json`.

Two details worth a look:

- **Discovery is a glob**, so a widget that grows a self-check is picked up without editing the test. A separate case asserts the glob matched *something* — an empty glob would make the parametrised test vacuously green, which is the same invisible silence this is closing.
- **A missing `node` skips locally but fails under `CI`.** Plenty of work here is Python-only, so a hard requirement would be annoying; but skipping in CI would restore exactly the silence being removed. GitHub's `ubuntu-latest` image ships node, so the `CI` branch only fires on a real regression in the runner image. Happy to add `actions/setup-node` to the `pytest` job instead if you would rather pin it explicitly.

## Verification

- `pytest tests/test_widget_clamp_checks.py -v` — 4 passed (the discovery guard plus one case per widget).
- Mutation-checked: appending a deliberate `assert.equal(1, 2)` to `calendar_day/tests/clamp_check.mjs` turns it red with the node assertion text in the failure output, and reverting turns it green. The bridge reports real failures rather than swallowing the exit code.
- `ruff check` and `ruff format --check` clean. `mypy` runs on `app` only, so `tests/` is out of its scope.

The change is a single new test file, so it adds coverage without touching any existing behaviour.

### DIFF
--- a/tests/test_widget_clamp_checks.py
+++ b/tests/test_widget_clamp_checks.py
@@ -1,0 +1,70 @@
+"""Run each widget's ``clamp_check.mjs`` self-check under pytest.
+
+The calendar widgets ship plain-node assertion scripts covering the client-side
+clamping helpers — ``clampScale``, ``clampToDay``, ``computeRange`` and friends.
+Their own headers say ``Run: node tests/clamp_check.mjs``, and nothing did:
+no CI step, no pytest bridge, no ``package.json`` test script referenced them.
+
+That is a coverage hole with a history. ``calendar_day``'s header records it:
+the cases "previously covered by test_smoke.py's
+test_scale_sliders_clamp_to_bounds" moved out of pytest when the clamp moved
+from ``server.py`` to ``client.js``, and the assertions went with them. They
+have been unrun since — free to drift from ``client.js`` with nothing to
+notice (#214).
+
+This is the "fold them into the Python smoke test via a subprocess call" option
+from that issue: the scripts stay where they are, next to the code they assert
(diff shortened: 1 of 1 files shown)


dmellok:
Merged, in v0.375.0. I'll stay with the runner-image node rather than an explicit setup-node step; your fail-loudly-in-CI branch already covers the case where that assumption breaks, which is the property that matters. Thanks for closing this one out.

Issue #257, opened 2026-08-24
list_widgets catalog: fields/q filter on /api/mcp/catalog


dmellok:
`list_widgets` returns the whole widget catalog in one response, and it is by
some distance the most expensive thing an MCP agent does. The context-cost
readout added in v0.359.0 made this visible: in a representative 28-step build
(create page, bind, probe, stream a code element, render, fix, push), the
catalog read accounted for roughly 70% of the entire build's context cost.

Measured against the bundled widget set (36 widgets):

| part of the response | bytes | ≈tokens | share |
| --- | --- | --- | --- |
| `widgets` | 19,325 | 4,831 | 84% |
| `appearance` | 3,354 | 838 | 14% |
| `libraries` | 1,113 | 278 | 4% |
| `icons` | 246 | 61 | 1% |
| **total** | **22,765** | **5,691** | |

Inside the `widgets` block, `desc` is 9,380 B (~2,345 tokens), 48% of it. The
longest single description is 821 characters. Everything else per widget
(`key`, `name`, `icon`, `fragments`, the `updates_on_*` flags) is small: the
median whole entry is 485 B.

That share only grows. Every widget installed from the marketplace adds its
description to every future `list_widgets` call, so an install with 60 widgets
pays roughly double, on every build, forever.

### Why it can't just be trimmed

The descriptions are what let an agent pick the right widget, so cutting them
blindly trades context cost for worse builds. They also aren't available
anywhere else right now: `get_widget_options(key)` returns `{key, options}`
only, with no `desc`, so the bulk list is currently the sole source.

### Options

1. **Move the long text, keep a short line.** A one-sentence summary in
   `list_widgets`, full `desc` returned by `get_widget_options(key)`. That fits
   the loop the tool docs already prescribe (list, then read options for the
   widget being placed), so the full text still reaches the agent for the two
   or three widgets it actually uses instead of all 36. Rough saving: ~2,000
   tokens per build on the bundled set.
2. **Let the caller ask for less.** `?fields=` or a `?q=` filter on
   `/api/mcp/catalog`, so an agent that knows it wants weather can say so.
   Cheap to add, but only helps agents that think to use it.
3. **Split the payload.** `appearance` (838 tokens of themes, styles and fonts)
   is only needed when something is actually being restyled, and could be its
   own tool rather than riding along with every catalog read.

(1) looks like the best value for the change involved, possibly with (3).

### Notes

- Whatever changes, the bridge stays as-is unless a tool description changes:
[...]

PR #269, opened 2026-08-29
feat(mcp): summarise widget descriptions in /catalog, serve the full text per widget

vaibhav8a:
Refs #257 — option **1**, the one you rate best value.

## The change

`/api/mcp/catalog` carries the **first sentence** of each widget's description, capped at 160 characters on a word boundary. `GET /widgets/<key>/options` now returns the full `desc` alongside the options.

That matches the loop the tool docs already prescribe — list, then read options for the widget being placed — so the full text still reaches the agent for the two or three widgets it actually uses instead of all 36.

## Measured, not estimated

On the bundled set, the `desc` block goes **9,256 B → 2,884 B**: a **69% cut**, roughly **1,593 tokens** off every `list_widgets` call. Slightly under your ~2,000 estimate, since a few descriptions are already one short sentence and save nothing.

The summaries read as intended — they are the sentence the author wrote first, not a truncation mid-thought:

| widget | full | summary |
| --- | --- | --- |
| `webpage` | 821 B | *Embed an external URL in a cell.* (32 B) |
| `tesserae_status` | 495 B | *Dashboard status strip: identity on the left, ambient device telemetry on the right.* (84 B) |
| `ha_locks` | 415 B | *A security-focused overview of every `lock.*` entity plus door / window / garage / opening binary_sensors.* (104 B) |

## Two decisions worth checking

**At the MCP layer, not in `catalog_entry`.** The editor palette reads the same builder and wants the full text, so summarising upstream would have changed the UI to save an agent's tokens. The trim sits next to the existing `sample`-stripping in `mcp_api.py`.

**A first-sentence split, not a sentence tokenizer.** These are hand-written single paragraphs; the cost of getting one slightly wrong is a slightly long summary, not a wrong answer. Abbreviations like "e.g." would split early — none in the bundled set do, and the fallback is graceful.

I did **not** touch `appearance` (option 3). It is a separate 838-token question about when a restyle is happening, and worth its own change.

## On "measure build quality, not just tokens"

Fair, and I can't measure that from outside — I have no build corpus to run before/after. What I can do is make the failure mode cheap: the full text is one call away on a tool the agent already uses, so if a summary turns out to be too thin for picking, the recovery is a call it was going to make anyway. If you would rather see a `?full=true` escape hatch on `/catalog` while you evaluate, say so and I will add it.

## Tests

Five added to `tests/test_mcp_api.py`, all five red without the change:

- every catalog `desc` is one sentence, no embedded newlines
- **the summarised total is under half the full total** — this is the one that guards the *point*; a summariser that returned the whole string would pass every other assertion here
- the full text round-trips: for the longest-described widget, `/options` returns exactly the manifest text, it is longer than the catalog line, and the summary is a genuine **prefix** of it rather than a rewrite
- an empty or whitespace-only description summarises to `""`
- a single 500-character sentence is cut at a word boundary with an ellipsis, never mid-word

Verified: `test_mcp_api.py` + the other two MCP suites 118 passed; `-k "catalog or widget_options or panels_schema"` 47 passed; `ruff check` and `ruff format --check` clean across `app` and `tests`. `mypy app` reports two pre-existing errors in `send_routes.py` and `rest_api.py` from a local werkzeug version skew — neither is in a file this touches.

### DIFF
--- a/app/mcp_api.py
+++ b/app/mcp_api.py
@@ -352,6 +352,38 @@ def _invalid_icons(layout: CanvasLayout) -> list[dict[str, Any]]:
 
 # -- catalog / widget options -------------------------------------------
 
+# Longest summary the catalog carries before it is cut at a word boundary.
+# Sized off the bundled set: most first sentences are well under this, and the
+# few that run past carry a second clause after a colon rather than being one
+# genuinely long sentence.
+_DESC_SUMMARY_MAX = 160
+
+# A sentence ends at . ! or ? followed by whitespace. Deliberately not a real
+# sentence tokenizer: these are hand-written single paragraphs, and the cost of
+# getting one slightly wrong is a slightly long summary, not a wrong answer.
+_SENTENCE_END = re.compile(r"(?<=[.!?])\s")
+
+
+def _desc_summary(desc: str) -> str:
+    """First sentence of a widget description, capped at a word boundary.
+
+    ``/catalog`` used to carry every widget's full ``desc``: 9,256 B across the
+    36 bundled widgets, 48% of the ``widgets`` block and the largest single
+    line item in an agent's context budget — paid on every build, for every
+    widget, whether or not it was placed (#257). Summarised it is 2,884 B.
+
+    The full text is not lost, it moves one call along: ``GET
+    /widgets/<key>/options`` now returns it, and that is the call an agent
+    already makes for the two or three widgets it actually places.
+    """
+    text = " ".join(desc.split())
+    if not text:
+        return ""
+    first = _SENTENCE_END.split(text, 1)[0]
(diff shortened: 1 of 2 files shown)


dmellok:
Merged, in v0.375.0, and it surfaced a follow-up worth knowing about: the tesserae-mcp bridge's client-side catalog summary was keying on "description", but the catalog field is named "desc", so the bridge's default list_widgets() view had been dropping descriptions entirely. Fixed in bridge 0.15.0 (on PyPI now), so your one-sentence desc is what agents actually see, and the served tool docs now describe the summary/full split. I'll pass on the ?full=true escape hatch; the full text being one get_widget_options call away seems like enough recovery, and it can be added later if a summary proves too thin in practice.

Issue #248, opened 2026-08-21
[Bug]: day calendar widget shows wrong content


bablokb:
### Tesserae version

v0.326.2 

### How is Tesserae running?

Other (describe in steps)

### Operating system

Debian Trixie on MicroCloud

### Panel hardware

na

### Steps to reproduce

1. Create a dashboard with two  cells: day and week
2. Render the dashboard with one entry for today
3. Move the calendar entry to tomorrow
4. Re-render the dashboard

This is what I get: I still have the event on Friday. It stretches for the whole day although the content correctly states it is from 10-11. The week view shows the correct story: nothing on Friday. It seems that the event for today is locally cached and not updated for the day-widget.

<img width="631" height="612" alt="Image" src="https://github.com/user-attachments/assets/93c9dc60-3c31-4167-aa2f-9dd26105716e" />

### Server logs

```shell

```

### Before submitting

- [x] I searched [existing issues](https://github.com/dmellok/tesserae/issues?q=is%3Aissue) and didn't find a duplicate.
- [ ] If I could, I tried reproducing on the latest minor release.

dmellok:
Not a caching bug, though the symptoms read exactly like one. Both halves of what you saw come from a single cause.

The day widget fetches a rolling window from *now*, `hours_ahead` wide, defaulting to 24. All-day events were filtered to the day being drawn; timed events never were. So with the default the window reaches into tomorrow, and your event moved to tomorrow 10-11 stayed in today's list.

The client then draws it against today's 0-24 axis. Its clamp exists for events that genuinely overlap today, an overnight shift or a multi-day trip, and the rule is "if neither edge is today, this runs straight through today, so draw the full column". An event living entirely on tomorrow trips that rule, which is your full-day block. The label prints the event's own times, which is why it still correctly said 10-11.

That also explains why the week view was right: it buckets by date, so it correctly showed nothing on Friday. The two widgets were not disagreeing about the data, only about which day the event belonged to.

Fixed at both ends in v0.334.0, on edge now. The server filters timed events to the day being drawn, and the client refuses to clamp an event that does not touch that day at all. Overnight and multi-day events that really do occupy today are unaffected, which the tests cover along with your exact reproduction.

One consequence worth knowing: `hours_ahead` above 24 no longer shows anything extra, since the widget draws a single day's axis and anything past it is filtered out.

[...]

PR #273, opened 2026-09-01
fix(calendar_day): open the day window at local midnight, not at now

### LINKED ISSUE #252: calendar_day: hours_ahead has no effect above 24, and the view loses the morning
Two related oddities in the day widget's time window, both surfaced while fixing #248.

## 1. `hours_ahead` above 24 does nothing

The widget draws a single day's 0-24 axis. Since #248, timed events are filtered to the day being drawn, so anything the wider window pulls in is discarded. The option is labelled "Hours of agenda to show" and silently stops having any visible effect past 24.

Either the label should say what it does (it narrows within today), or the widget should be able to draw more than one day when asked for more than one day.

## 2. Events earlier today are never loaded

The window starts at `now`, not at local midnight:

```python
now = now_local.astimezone(UTC)
end = now + timedelta(hours=hours_ahead)
```

So as the day progresses the day view quietly loses the morning: at 14:00 a 10:00-11:00 meeting is not in the payload at all, and the auto-fitted range shrinks around whatever is left. For a widget that draws a day's timeline that reads as missing data rather than as an agenda.

Whether that is intended ("agenda ahead") or a bug depends on what the widget is meant to be. Starting the window at local midnight and keeping `hours_ahead` as a forward bound would make it a day view; leaving it is defensible if the intent is strictly "what is coming up".

Worth settling both together, since they are the same question about what the window means.

vaibhav8a:
## What

The day widget's fetch window now opens at local midnight. `hours_ahead` keeps its meaning as a forward bound.

Fixes #252, settling both halves.

## The decision the issue asks for

You framed it as one question — what does the window mean — and left it open.

**It is a day view.** The widget draws a single day on a 0-24 axis and, since #248, filters every event to that day. Nothing about it is an "agenda ahead" surface: the axis is the whole day whether or not there is anything in the morning.

So the window opens at local midnight, and `hours_ahead` becomes a purely forward bound measured from now.

## Part 2: the morning

```python
now = now_local.astimezone(UTC)
end = now + timedelta(hours=hours_ahead)
```

At 14:00 a 10:00-11:00 meeting was **not in the payload at all** — not filtered out downstream, never fetched. The auto-fitted range then shrank around what was left, so a day view drew an afternoon and gave no sign the morning existed.

The combination is what made it indefensible: a 2 h agenda at 09:00 showed 09:00-11:00 and called it a day. It now shows the day so far plus two hours.

## Part 1: the label

Nothing above 24 has a visible effect, because everything is filtered to the day being drawn. Rather than teach the widget to draw several days — which is a different widget, and out of scope here — the label now says what the option does:

> "Hours of agenda to show" → **"Hours ahead to include (within today)"**

If you would rather `calendar_day` grow a multi-day mode, that is worth its own issue and I am happy to take it; it changes the axis, the clamping rule and the auto-fit, so it did not belong in this fix.

## Tests

Four added, asserted through **the arguments passed to `calendar_core`**, not through the returned events. That matters: the test stub returns whatever it is given regardless of the window, so a test reading the output cannot see this bug at all. The window *is* the bug.

Two of the four are confirmed red before the change.

`test_calendar_day_uses_app_timezone_for_display_date` pinned `start == now` and is updated. Its intent is unchanged, and the assertion is now **stronger**: 18:30 UTC is 02:30 the next day in Hong Kong, so the day being drawn began at 16:00 UTC — while reading `now` back proved nothing about the zone, since `now` is the same instant everywhere.

## Gates

- [x] `ruff check .` — clean
- [x] `ruff format --check .` — clean (543 files)
- [x] `pytest -k calendar` — 62 passed
- [x] Full `pytest` run before this: 664 passed with only the timezone test above, which this PR updates

### DIFF
--- a/plugins/calendar_day/plugin.json
+++ b/plugins/calendar_day/plugin.json
@@ -18,7 +18,7 @@
     {
       "name": "hours_ahead",
       "type": "number",
-      "label": "Hours of agenda to show",
+      "label": "Hours ahead to include (within today)",
       "default": 24
     },
     {
--- a/plugins/calendar_day/server.py
+++ b/plugins/calendar_day/server.py
@@ -34,12 +34,23 @@ def fetch(
 
     zone = app_timezone()
     now_local = datetime.now(zone)
-    now = now_local.astimezone(UTC)
-    end = now + timedelta(hours=hours_ahead)
+    # Start at LOCAL MIDNIGHT, not at `now`.
+    #
+    # This widget draws a single day on a 0-24 axis and filters everything to that day
+    # (see below), so it is a day view. Starting the fetch at `now` meant it quietly lost
+    # the morning as the day went on: at 14:00 a 10:00-11:00 meeting was not in the
+    # payload at all, and the auto-fitted range then shrank around whatever was left. On
+    # a widget that draws a day's timeline that reads as missing data rather than as an
+    # agenda (#252).
+    day_start_local = now_local.replace(hour=0, minute=0, second=0, microsecond=0)
+    start = day_start_local.astimezone(UTC)
+    # `hours_ahead` keeps its meaning: a FORWARD bound, measured from now. It no longer
+    # decides where the window begins, only how far past this moment to look.
+    end = now_local.astimezone(UTC) + timedelta(hours=hours_ahead)
     try:
         events = core.server_module.load_events(
             feeds_filter,
-            now,
+            start,
             end,
             data_dir=Path(core.data_dir),
         )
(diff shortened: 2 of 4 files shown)


dmellok:
Merged. Agreed on the framing: the #248 filter already committed this widget to being a day view, so midnight is the only start that's consistent with it.

Asserting the window through the `load_events` arguments was the right call; the stub returns whatever it's given, so an output-level test couldn't have seen this. The reworked Hong Kong assertion is also genuinely stronger than the `start == now` it replaced.

One side effect worth noting: an overnight event that started yesterday and ends this morning was also invisible in the afternoon before this, and now shows correctly, so the fix covers slightly more than the issue described.

On a multi-day mode for `calendar_day`: agreed it's out of scope here, and I'm not sure it belongs in this widget rather than a variant. If you want to open that issue, go ahead.

This goes out in v0.380.1 on edge.

PR #281, opened 2026-09-06
fix(lineups): keep deleted-display lineups visible

charmmmz:
## What this changes

Keep Lineups whose targets are all missing visible in an **Unavailable displays** group, with their existing Edit and Delete actions. Lineups with a surviving target stay under that display; unassigned Lineups retain their existing group.

## Why

Deleting a display preserves its Lineups, even with associated-data cleanup enabled. The web list grouped records under registered displays or empty target lists, so a non-empty binding to a deleted display fell into neither group. Companion still received the same record from the API and showed it as unavailable, making this look like an app caching issue.

Reproduction: create a Lineup bound to a display, delete that display in Settings, then open Lineups. The retained record now stays accessible on the web.

## Test plan

- [x] Python 3.11: `pytest -n 4 -q` — 4,293 passed in 310.31s.
- [x] `mypy app` — 203 source files.
- [x] Nine regression cases cover all four intents with and without associated-data cleanup, edit-page access, deletion, inherited targets, and shared/unassigned grouping.

## Checklist

- [x] Regression tests added.
- [x] Changelog entry added under Unreleased.
- [x] Stored bindings and device-deletion policy preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,11 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- **Lineups remain visible after their displays are deleted.** Lineups whose
+  targets are all unavailable now appear in an **Unavailable displays** group
+  on the web, where they can still be edited or deleted. Existing Lineups are
+  preserved, including when device deletion also clears associated data.
+
 - **A touch action no longer repaints a different lineup page over the
   one on glass.** The post-action reconcile resolved "the page the
   device is showing" from the lineup's nav record before the live frame,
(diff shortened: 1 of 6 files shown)


dmellok:
Merged, thanks. Good catch on the lineups vanishing once their last display was gone; that path had no coverage at all.

I pushed one commit on top before merging. The "still editable" part had a gap: the deck editor's display select only lists registered devices, so opening a retained lineup and saving would clear its binding, and the schedule form would preselect the first dashboard once the wiped page dropped out of the list. Both editors now keep the stale value selected and labelled until you pick a replacement, and the test does a save round-trip to pin that down. I also trimmed the changelog line, since the wipe path still deletes the lineup's pages even though the deck record survives.

Two things I left for a follow-up rather than this PR: the scheduler still fires enabled lineups bound to a deleted display (and logs a failed push each tick), and the Companion lineup projection has no marker for a missing device, so the app and the web will disagree on the same record. If you have a view on how the app should surface that, worth a thread before either of us builds it.

PR #284, opened 2026-09-06
fix(quantizer): LAB matching sees the diffused error (#202)

### LINKED ISSUE #202: LAB colour matching silently disables dithering
Selecting **LAB** or **chroma-aware** colour matching in the Calibration tab turns error diffusion off. The frame comes back posterised, with no dither texture at all, on every panel.

Found by @varanu5 while calibrating a BWRY panel in #201, and it is not specific to that gamut or to any one dither algorithm.

### Cause

`_error_diffusion` builds `buf_l` / `buf_a` / `buf_b_lab` once from the original image (`app/quantizer.py:1132`) and never writes diffused error back into them: only `buf_r` / `buf_g` / `buf_b` receive it (`app/quantizer.py:1200`). The LAB branch then picks the nearest ink by reading those stale buffers (`app/quantizer.py:1165`). Every pixel is judged against its pristine value, so the diffused error is ignored and the result is plain nearest-colour quantisation.

`_error_diffusion` never sees a gamut, and `atkinson` / `jarvis` / `stucki` all call it directly (`floyd-steinberg` detours into it when LAB is selected), so this affects every gamut and every error-diffusion dither.

### Signature

A flat 40x40 mid-grey patch, counting distinct inks in the output:

| gamut | rgb | lab / chroma-aware |
|---|---|---|
| `bwry_4` (4 inks) | 4 | 1 |
| `waveshare_e6` (6 inks) | 6 | 1 |
| `inky_7colour` (7 inks) | 5-7 | 1 |

Measured across 12/12 gamut x dither pairings under both LAB modes.

### Fix

Recompute LAB per pixel after each error write, inside a Python hot loop shared by every panel, so the cost needs measuring before committing to an approach.

### Notes

`tests/test_bwry_calibration.py::test_lab_matching_still_loses_dither_feedback` is a tripwire asserting the current broken behaviour; it will start failing when this is fixed, which is the signal to invert it and to add a LAB-matching palette preset. No bundled profile selects LAB matching today.

vaibhav8a:
## What this changes

`_error_diffusion` now derives each pixel's LAB coordinates from the pixel **as it stands when it is matched**, instead of from a pre-pass over the original image.

Closes #202

## Why

The LAB buffers were built once from the pristine image (`quantizer.py:1132`) and never received the diffused error — only `buf_r/g/b` did (`:1200`). The LAB branch then read those stale buffers to pick the nearest ink (`:1165`), so every pixel was judged against its original value, the diffusion was ignored, and the result was plain nearest-colour quantisation.

The source-LAB pre-pass is gone. Only the palette is converted up front, because only the palette is fixed.

## The cost, measured

The issue asks for this before committing to an approach. sRGB→linear is a 256-entry table — exact rather than approximate, since `_srgb_to_lab` quantises to `uint8` before converting — and the conversion is memoised on the rounded RGB triple.

800×480, floyd-steinberg:

| input | before | after | |
|---|---|---|---|
| flat dashboard | 0.74s | 0.76s | **+3%** |
| photo noise | 0.73s | 0.96s | +32% |
| `rgb` match (control) | 0.65s | 0.66s | unchanged |

The memo pays off exactly where the workload is: a dashboard frame is mostly flat colour, so most pixels are a dict hit. Pure noise is the worst case and is not what these panels render. The `rgb` path is untouched.

I also checked the scalar conversion against `_srgb_to_lab` directly — same constants, same piecewise curve, max delta < 1e-3 across black, white, mid-grey, primaries and the toe region — so a pixel that received no error converts to the identical triple it did before.

## Test plan

- [x] `pytest -q` on the affected areas: `-k "quant or dither or palette or calibrat or render or panel or bin"` — **724 passed**, 10 skipped (Playwright not installed)
- [x] `pytest tests/test_bwry_calibration.py` — 40 passed
- [x] `ruff check . && ruff format --check .` on the changed files — clean

**The tripwire fired, as designed.** `test_lab_matching_still_loses_dither_feedback` said "if this test starts failing, the upstream bug was fixed", and it did. It is inverted to `test_lab_matching_keeps_dither_feedback`.

I added a second test with it, because counting inks is not enough on its own — a frame split into two solid halves would pass an ink count. `test_lab_matching_produces_an_alternating_pattern_not_just_two_inks` asserts a flat row actually alternates. On unfixed code it reports `only 0 ink changes across a 40px flat row`.

The measured signature, flat 40×40 mid-grey, distinct inks:

| gamut | rgb | lab / chroma-aware (before) | lab / chroma-aware (after) |
|---|---|---|---|
| `bwry_4` | 3 | 1 | 2 |
| `waveshare_e6` | 2 | 1 | 2 |
| `inky_7colour` | 5–7 | 1 | 2 |

Worth saying plainly: LAB lands on fewer inks than `rgb` here, and I believe that is correct rather than a residual bug. For a neutral grey, a perceptual match picks the two nearest *neutral* inks; `rgb`'s euclidean match wanders into chromatic ones. The output is a clean checkerboard (`1010101…`) where before it was a solid block. On a gradient both matchers use the same number of inks.

## Not in this PR

`test_no_bundled_profile_selects_lab_matching` is left alone, so no shipped preset steers users onto LAB. The issue says a LAB preset "becomes worth reconsidering" once this is fixed — that is a palette decision with visual consequences, and it seemed better to leave it to you than to bundle it in a correctness fix.

### DIFF
--- a/app/quantizer.py
+++ b/app/quantizer.py
@@ -1211,6 +1211,45 @@ def _native_colour_mask(
     return flat
 
 
+#: sRGB transfer curve for the 256 integer channel values, precomputed.
+#: :func:`_srgb_to_lab` quantises to ``uint8`` before converting, so a table
+#: indexed by the rounded channel is exact rather than an approximation of it.
+_SRGB_LINEAR_8: tuple[float, ...] = tuple(
+    (v / 12.92) if (v := i / 255.0) <= 0.04045 else ((v + 0.055) / 1.055) ** 2.4 for i in range(256)
+)
+
+#: D65 reference white, and the CIELAB toe constants, unpacked once so the
+#: per-pixel path does no attribute lookups.
+_LAB_REF_X = 0.95047
+_LAB_REF_Y = 1.00000
+_LAB_REF_Z = 1.08883
+_LAB_DELTA = 6.0 / 29.0
+_LAB_DELTA_CUBED = _LAB_DELTA**3
+_LAB_TOE_SCALE = 1.0 / (3 * _LAB_DELTA**2)
+_LAB_TOE_OFFSET = 4.0 / 29.0
+
+
+def _lab_of_rgb8(r8: int, g8: int, b8: int) -> tuple[float, float, float]:
+    """Scalar sRGB (0-255 ints) -> CIE L*a*b*, matching :func:`_srgb_to_lab`.
+
+    The vectorised version is the right tool for a whole image and the wrong
+    one for a single pixel: error diffusion has to re-read a pixel *after* its
+    neighbours have written into it, so the conversion has to happen inside
+    the sequential loop. Same constants and same piecewise curve, so a pixel
+    that never received any error converts to the identical triple.
+    """
+    lr = _SRGB_LINEAR_8[r8]
+    lg = _SRGB_LINEAR_8[g8]
+    lb = _SRGB_LINEAR_8[b8]
+    xn = (0.4124564 * lr + 0.3575761 * lg + 0.1804375 * lb) / _LAB_REF_X
+    yn = (0.2126729 * lr + 0.7151522 * lg + 0.0721750 * lb) / _LAB_REF_Y
+    zn = (0.0193339 * lr + 0.1191920 * lg + 0.9503041 * lb) / _LAB_REF_Z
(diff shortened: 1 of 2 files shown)


dmellok:
Merged, thanks. The timing table answered the question the issue asked, and the alternating-row test is the one that would have caught this originally. One change on top: I capped the memo at 64k entries, since a photo's diffused values spread over far more triples than a dashboard's and an unbounded dict on a large panel could grow to one entry per pixel.

PR #285, opened 2026-09-06
fix(gallery): the sequential cursor is per device (#209)

### LINKED ISSUE #209: Gallery sequential mode advances on preview and probe renders, not just device paints
The Gallery widget's sequential mode keeps its position in a counter file that advances on every `fetch()` call (`plugins/picture_gallery/server.py:280-296`):

```python
next_idx = (current + 1) % len(images)
idx_file.write_text(str(next_idx), encoding="utf-8")
```

`fetch()` runs for any render, not just one that reaches a panel. So opening the dashboard editor, hovering a card for the list preview, or probing the widget all step the album forward, and the next real paint skips whatever those renders consumed.

The cursor is also keyed by folder and orientation only, so it is global rather than per page, per device, or per cell. Two dashboards pointing at the same folder share one position and both advance it.

Reported on Discord alongside a related cadence problem: a panel sleeping 30 minutes against a dashboard refreshing every 15 sees every second picture, because the server steps the gallery twice per wake and only the newest render survives in the latest-wins slot. The cadence half is fixable by the operator (and easier now that the Updates control takes any interval); this half is not.

### Possible directions

- Advance only for renders that are actually destined for a device, leaving previews and probes on the current image. Needs `fetch()` to know why it is being called, which no widget currently does.
- Or move the cursor out of the widget's data dir and key it per device, so each panel walks the album at its own pace and previews are naturally separate.

The second is the better model but is a larger change, since widget state is currently folder-scoped by construction.

vaibhav8a:
## What this changes

The Gallery's sequential cursor is now keyed by device, so each panel walks the album at its own pace.

Closes #209

## Why

The preview guard already landed and stops a hover or a probe consuming a photo. The other half of the issue is untouched: the cursor was keyed by folder and orientation alone, so it was global. Two dashboards pointing at one folder shared a position and each advanced it, and a panel woken between the other's renders skipped whatever they had consumed.

The issue calls a per-device cursor "the better model but a larger change, since widget state is currently folder-scoped by construction". It turns out the mechanism already exists: the composer hands `target_device_id` to any widget that declares `render.per_device_id` in its manifest (`composer.py:724`), which is how `tesserae_status` already works. The gallery opts in and keys the cursor file with it, so this stays inside the existing model rather than inventing a second one.

**A render with no device id keeps the folder-scoped file it has always used.** An unbound or virtual-panel render has no device to walk for, and existing albums keep their position across the upgrade instead of resetting to the first photo.

**The device id is hashed, not sanitised.** It is operator-supplied and it lands in a path. A substitution scheme has to be audited for what it lets through; a hex digest has nothing to let through.

## Test plan

- [x] `pytest plugins/picture_gallery/tests/ -q` — 13 passed
- [x] `ruff check . && ruff format --check .` on the plugin — clean

Four new tests:

- **two devices walk independently** — each sees the whole album from its own position
- **one device's paints do not move another's** — interleaves two panels and asserts neither skips, which is the failure an operator actually reports
- **an unbound render keeps the shared cursor** — and a device render then writes a *second*, differently-named file, since "kept the old name" is only meaningful against something that changes it
- **a device id never reaches the filename raw** — drives `../../etc/passwd` through as a device id and asserts every cursor file stays inside the widget's data dir

One thing worth flagging from writing those: my first version of the unbound-cursor test asserted the file was named `.sequential_index_`, and it failed — the real name is `.sequential_index__root`, because the root folder segment is `_root`. The behaviour was right and my assertion was wrong. It now asserts the property (one file, no device suffix, still present after a device render) rather than a literal filename that encodes an unrelated detail.

**The manifest flag changes the push pipeline for every page with a gallery cell.** `render.per_device_id` is not only a "hand me the device id" switch. `_page_needs_per_device_render` (app/push.py) turns it into one compose + Playwright capture per bound device instead of one per panel group, for the whole page, in every mode. A gallery page bound to three same-size panels now does three captures, every other widget on that page fetches three times, and the captures run back to back so the last device's frame lands later inside the smart-sync lead window. Random mode pays that without reading the id, and same-panel devices in random mode no longer get the same photo. Accepted for this PR because the fan-out is the mechanism the fix needs and the flag is not option-aware; copy the flag into another manifest only if that cost is wanted.

## Maintainer follow-up (merged with fixes)

- The per-device cursor file is named after the device (`.sequential_index_<folder>_<device>`) when the id matches `DEVICE_ID_RE`; anything else is still hashed.
- A device's first render seeds from the folder-scoped cursor, so existing albums keep their position after the upgrade instead of restarting at photo 0.
- The independence test asserts the interleaved sequence literally, so it fails against the old code.

## Not in this PR

The cadence half mentioned in the issue — a panel sleeping 30 minutes against a dashboard refreshing every 15 — is operator-fixable and unchanged here.


dmellok (changes requested):
Thanks for picking up the second half of #209. The per-device key is the right model and the interleaving test is the one that matters. Two things I'd like fixed before merge, two optional.

**1. The manifest flag changes the push pipeline for every page with a gallery cell, and the description should say so.**

`render.per_device_id` isn't a "hand me the device id" switch. `_page_needs_per_device_render` (app/push.py:174) turns it into one compose + Playwright capture per bound device instead of one per panel group, for the whole page, in every mode. So a gallery page bound to three same-size panels now does three captures, every other widget on that page fetches three times, and the captures run back to back so the last device's frame lands later inside the smart-sync lead window. Random mode pays all of that for nothing, since it never reads the id. Same-panel devices also stop getting the same random photo.

I'm fine accepting that cost for this PR, because per-device fan-out is the mechanism the fix needs and the flag isn't option-aware today. But the description currently reads as if this stays inside the widget, and it doesn't. Please add a paragraph on the fan-out consequence so the next person who copies this manifest knows what they're opting into.

**2. Existing albums do restart at photo 0 after the upgrade.**

The "unbound render keeps the shared cursor" guarantee only covers the path with no device id. Binding is the delivery model, so every real push now arrives with `?device_id=`, opens a new per-device file, misses, and starts at -1. The folder-scoped file is never read again on that path.

Suggested fix: on a per-device miss, seed `current` from the folder-scoped file before falling back to -1. Three lines, and the docstring on `test_a_render_with_no_device_keeps_the_shared_cursor` becomes true.

**3. (optional) `test_two_devices_walk_the_album_independently` passes against the old code.**

With three images and three renders per device, the global cursor gives kitchen a,b,c, wraps, then hallway a,b,c, so `sorted()` and `kitchen == hallway` all hold. I checked by reverting server.py to main: only the interleaving test and the shared-cursor test fail. Either drop it or assert the interleaved sequence literally (kitchen, hallway, hallway, kitchen should give a, a, b, b). The traversal test also passes on main for the same reason.

**4. (optional) The hash could be a plain suffix.**

Device ids are validated at creation against `DEVICE_ID_RE` (`^[a-z][a-z0-9_-]{1,31}$`, app/device_service.py:37), and push only fans out to registry ids, so nothing outside that alphabet reaches `fetch()`. The same filename already carries `folder_segment` raw on the strength of a regex. A readable `.sequential_index_trip_kitchen` means an operator can delete one panel's cursor to restart its album, which they can't do with a digest. Not blocking if you'd rather keep the hash.

Things I noticed that predate this PR and are out of scope here: cursor writes are non-atomic, deck warms and touch prewarms advance the cursor for frames that may never be promoted, and deleting a folder leaves its cursor files behind.

Issue #164, opened 2026-07-30
[Bug]: Schedule page not updating


bablokb:
### Tesserae version

v0.213.2

### How is Tesserae running?

Other (describe in steps)

### Operating system

Debian Trixie on MicroCloud

### Panel hardware

N/A

### Client repo + version

N/A

### Steps to reproduce

Open the "Schedules" from the main menu. The "Next 24 hours" box is strange. I don't know what the horizontal line from "Now" to (about) 23:00 is trying to tell me. In addition, "Now" does not  update. I did not define any quiet hours, so it is not clear why the line stops suddenly.

<img width="1412" height="536" alt="Image" src="https://github.com/user-attachments/assets/885dfc16-dc01-464f-be94-00d1fd6e0532" />

### Server logs

```shell

```

### Before submitting

- [x] I searched [existing issues](https://github.com/dmellok/tesserae/issues?q=is%3Aissue) and didn't find a duplicate.
- [ ] If I could, I tried reproducing on the latest minor release.

dmellok:
Thanks, this is a couple of real issues plus a labeling gap.

The box stacks three things that aren't labeled well: a vertical "Now" line, one horizontal lane per schedule, and a tick on that lane for each projected fire. So the horizontal line you're seeing is your schedule's lane with its fire ticks, and the vertical accent is "Now."

Two things are genuinely wrong:

- **"Now" doesn't move.** It's computed once when the page renders and there's no live update, so it only advances on reload. It shouldn't look like a live clock when it isn't; I'll either tick it client-side or label it as the page-load time.
- **The ticks stopping at ~23:00 isn't quiet hours, it's a draw cap.** Your 3-minute schedule fires ~480 times over 24 hours, but the timeline only draws the first ~200 ticks, so they run out roughly ten hours in and just stop with no indication. Quiet hours aren't drawn on this view at all, which is why it made no sense against your config. I'll raise the cap and/or thin dense schedules to a density band so the lane reads correctly across the full window.

Broader context: schedules, decks, and rotations have grown into three overlapping ways to say "show these dashboards over time," and the seams between them are confusing (this timeline being one of them). I'm planning to unify them into one model rather than patch each separately, so some of this rough edge will get resolved as part of that. The two concrete bugs above I'll fix regardless.

bablokb:
> so it only advances on reload.

No, this isn't the case. I tried reloading the page or switching to other pages but that did not help.

dmellok:
[...]

PR #286, opened 2026-09-06
fix(rail): thin dense schedules instead of truncating them (#166)

### LINKED ISSUE #166: Schedules timeline: dense schedules truncate silently (~200-tick draw cap)
The "Next 24 hours" timeline silently truncates dense schedules, so the fire ticks stop partway across the window with no indication, reading as "the schedule stopped firing."

Two caps in app/schedule_routes.py:
- `_project_fires` stops after `safety < 500` iterations.
- `_build_timeline` then draws only `fires[:200]`.

Concrete case from #164: a 3-minute interval schedule projects ~480 fires over 24h, but only the first ~200 ticks are drawn, so they run out roughly ten hours in and stop. The per-row `(count)` label still shows the larger pre-cap number, so the label and the drawn ticks disagree.

### Fix direction
Raise or remove the draw cap, or thin dense schedules into a density band / heat strip instead of discrete ticks so the lane reads correctly across the full 24h regardless of interval. Keep the `(count)` label consistent with what's drawn.

vaibhav8a:
## What this changes

The 24h rail thins a dense schedule across the whole day instead of drawing only its first few dozen fires.

Closes #166

## Why

The rail drew `marks[:48]`. A 3-minute schedule projects ~480 fires a day, so the ticks ran out about two hours in and the lane read as "the schedule stopped firing" — while the `refreshes today` label beside it still showed the full 480, so the label and the ticks disagreed.

Taking every *n*th mark keeps the lane spanning the window it covers, which is what a rail is for: the reader is judging *when* a schedule fires and roughly how densely, not counting ticks. The count label stays the honest total and is now consistent with a lane that reaches the end of the day.

First and last marks are always kept, so the lane starts and ends where the schedule does, and duplicates are dropped so two ticks never stack on one pixel and understate density.

## One correction to the issue

The issue reports this against `_build_timeline` in `app/schedule_routes.py`, with its `fires[:200]` cap. **That function has no production caller** — a repo-wide search finds only `tests/test_schedule_routes.py` importing it.

The rail an operator actually sees is built in `app/deck_routes.py:479`, and its cap is **48**, not 200. So the truncation bites about four times sooner than the issue estimates: for the 3-minute case in #164, the ticks stop at 9.8% of the rail — roughly 02:20 on a day that runs to 24:00.

I fixed the live one. `_build_timeline` is left alone; if it is dead it wants deleting rather than patching, and that is your call rather than something to fold into a bug fix.

## Test plan

- [x] `pytest tests/test_deck_rail_thinning.py -q` — 7 passed
- [x] `pytest -q -k "deck or schedule or rotation or rail"` — 422 passed, 5 failed
- [x] `ruff check . && ruff format --check .` on the changed files — clean

**On those 5 failures:** they are all in `tests/test_rotation_scheduler.py` and they fail identically on `main` with my change stashed. Not caused by this PR, and I have not touched them.

The tests assert the properties rather than a mark count: that a sparse schedule is untouched, that a dense lane still reaches >99% of the day, that the cap is never exceeded at any density, that marks stay ordered and unique, and that ticks appear in every quarter of the day — the last one being what head-truncation could never satisfy.

### DIFF
--- a/app/deck_routes.py
+++ b/app/deck_routes.py
@@ -249,6 +249,43 @@ def _cycle_fires_on(rotation: Any, day_start: datetime) -> list[datetime]:
     return fires
 
 
+#: Ticks the 24h rail draws. A rail is a few hundred pixels wide, so past
+#: this the marks stop being separable and only cost DOM.
+MAX_RAIL_MARKS = 48
+
+
+def _thin_marks(marks: list[float]) -> list[float]:
+    """Reduce *marks* to at most :data:`MAX_RAIL_MARKS`, spread across the day.
+
+    The rail used to draw ``marks[:48]``. A 3-minute schedule projects ~480
+    fires, so the ticks ran out about two hours in and the lane read as "the
+    schedule stopped firing" -- while the ``refreshes today`` label beside it
+    still showed the full count, so the two disagreed (#166).
+
+    Taking every *n*th mark instead keeps the lane spanning the window it
+    covers, which is what the rail is for: the reader is judging *when* the
+    schedule fires and roughly how densely, not counting ticks. The count
+    label remains the honest total, and it is now consistent with a lane that
+    reaches the end of the day.
+
+    The first and last marks are always kept, so the lane starts and ends
+    where the schedule does.
+    """
+    if len(marks) <= MAX_RAIL_MARKS:
+        return marks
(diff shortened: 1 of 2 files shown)


dmellok:
Merged, thanks. You were right that `_build_timeline` has no production caller; I'll deal with deleting it separately. The CI failure on the branch was an unrelated gtfs test, not your change.

Issue #170, opened 2026-07-30
[Bug]: System does not seem to respect system time


pktiuk:
### Tesserae version

0.212.0

### How is Tesserae running?

Docker compose

### Operating system

Ubuntu 24.04

### Steps to reproduce

1. Set timezone
2. Check time in this timezone
3. Compare it against `Next 24 hours widget`/`recent logs`

### Expected behaviour

Times should match or be similar

### Actual behaviour

Time seems to be around 1 hour behind
<img width="1551" height="1188" alt="Image" src="https://github.com/user-attachments/assets/309ef71d-8311-4e35-bd38-7cf22daf6081" />

In logs I also saw:

<img width="863" height="271" alt="Image" src="https://github.com/user-attachments/assets/07011504-f5ea-43ca-8e37-da1f0e296291" />

(my clock time was 19:01)

### Server logs

```shell

```

### Before submitting

- [x] I searched [existing issues](https://github.com/dmellok/tesserae/issues?q=is%3Aissue) and didn't find a duplicate.
- [ ] If I could, I tried reproducing on the latest minor release.

pktiuk:
At first glance this issue seemed to be similar to https://github.com/dmellok/tesserae/issues/164, but that one is more about the interface not the actual time

pktiuk:
Oh, I see that this issue may be a duplicate of https://github.com/dmellok/tesserae/issues/143 (still not 100% sure)

dmellok:
Confirmed, and it's the same root cause as #164: the "Next 24 hours" widget and the last-fired times were rendered in the server's container timezone (UTC on Docker) rather than your configured app timezone, so they read your offset behind. Fixed in v0.213.5.

It's the same class of bug as #143 but in the schedule views specifically. The events log page already used your configured timezone; if you still see a wrong timestamp anywhere after upgrading, flag which view and I'll chase it.

pktiuk:
I will check this log page case again tomorrow (just to be sure).

pktiuk:
Sorry, I forgot to do it.

> Confirmed, and it's the same root cause as https://github.com/dmellok/tesserae/issues/164: the "Next 24 hours" widget and the last-fired times were rendered in the server's container timezone (UTC on Docker) rather than your configured app timezone, so they read your offset behind. Fixed in v0.213.5.

You are right.

Thank you for fixing it.

Issue #143, opened 2026-07-27
[Bug]: /status: server_time is wrong


bablokb:
### Tesserae version

v0.202.0

### How is Tesserae running?

Other (describe in steps)

### Operating system

Debian Trixie on MicroCloud

### Panel hardware

n/a

### Client repo + version

n/a

### Steps to reproduce

Use /status endpoint. The endpoint returns `server_time` and `local_time` (formatted). While `local_time` is correct, `server_time` is off by a couple of minutes (plus TZ offset of course).

The client log shows:
  - date-header of the server matches the local_time (plus TZ offset): `08:46:09`
  - server_time is `1785141760` (the dump of the json does not show this, but printing it with enough width at the very end gives the number)
  - this translates to `08:42:40`

In another test, the server_time was ahead of the local_time by a few minutes.
 
```
Tesserae: posting to http://tesserae-dev.bablokb-local.de:8765/api/v1/device/cp_magtag/status
Tesserae: {
Tesserae:   battery_mv: 4258.0
Tesserae: }
Tesserae: api-response: code=200
Tesserae: headers:
Tesserae: {
Tesserae:   access-control-max-age: 86400
Tesserae:   access-control-allow-headers: Authorization, Content-Type, If-None-Match, X-Tesserae-Token, X-Pairing-Code
Tesserae:   server: tesserae
Tesserae:   content-type: application/json
Tesserae:   content-length: 270
Tesserae:   access-control-allow-origin: *
Tesserae:   date: Mon, 27 Jul 2026 08:46:09 GMT
Tesserae:   access-control-allow-methods: GET, POST, OPTIONS
Tesserae:   access-control-expose-headers: ETag, Content-Location
Tesserae: }
Tesserae: {
Tesserae:   tz_offset_seconds: 7200
Tesserae:   status: 200
Tesserae:   server_time: 1.78514e+09
Tesserae:   local_time: 2026-07-27T10:46:09+02:00
Tesserae:   tz: Europe/Berlin
Tesserae:   dst_active: True
Tesserae:   next_poll_s: 60
Tesserae:   config: {'sleep_interval_s': 60, 'always_on': False, 'palette_profile_slug': 'paperlesspaper-spectra6'}
Tesserae: }

...
ui_application: updating RTCs to: 1785141760.0000
ui_application: now: 2026-07-27 08:42:40
```

### Server logs

```shell

```

### Before submitting

[...]

PR #287, opened 2026-09-06
fix(rail): the now-marker ticks in the configured timezone (#165)

### LINKED ISSUE #165: Schedules timeline: "Now" marker is a static snapshot, doesn't advance
The "Next 24 hours" timeline on the Schedules page renders a "Now" marker (vertical accent line + label) whose position is computed once, server-side, at page render (`now_pct` from `datetime.now()` in `_build_timeline`, app/schedule_routes.py). There is no client-side update, so the marker is frozen at page-load time and only advances on reload.

The accent styling reads as a live clock, which it isn't. Reported in #164.

### Fix options
- Advance the marker client-side (recompute `left` from wall-clock on an interval), or
- Relabel it as the page-load time so it doesn't imply liveness.

Client-side ticking is the better UX; it only needs the window start/end already available to the template.

vaibhav8a:
## What this changes

The rail's now-marker advances in `settings.app.timezone`, not in whatever zone the operator's browser happens to be in.

Closes #165

## Why — and a correction to the issue

#165 reports the marker as a static snapshot that never advances, against `_build_timeline` in `app/schedule_routes.py`.

That function has no production caller (only a test imports it), and the rail an operator actually sees — built in `deck_routes.py`, drawn by `decks.html` — has ticked client-side since the Lineups redesign:

```js
const tick = () => {
  const now = new Date();
  const pct = ((now.getHours() * 60 + now.getMinutes()) / 1440) * 100;
  ...
setInterval(tick, 60000);
```

So the liveness half is already done. **What is left is that the tick reads the wrong clock.**

The server anchors `now_pct` to `datetime.now(app_timezone())` — deliberately, with a comment citing #164 / #170 / #143 about not inheriting the container's UTC. The client then recomputes from `new Date().getHours()`, the *browser's* zone. On load the marker is right; a minute later it jumps by the offset between the two. That is the same bug those issues fixed server-side, arriving again from the other side, and it is invisible to anyone whose browser and server share a zone.

## The fix

The page carries the IANA name, and the tick formats through `Intl.DateTimeFormat` with `timeZone` — so DST is the platform's problem rather than offset arithmetic I would have to get right. An unresolvable name falls back to the browser clock, which is the best guess available and exactly what happened before.

The interval is also aligned to the minute boundary rather than firing 60s after page load, so the marker moves when the clock does instead of on an offset of whenever the page opened.

## Test plan

- [x] `pytest tests/test_deck_rail_now_timezone.py -q` — 4 passed; all fail on `main` (`KeyError: 'now_tz_name'`)
- [x] `pytest -q -k "deck or decks or rail"` — 206 passed
- [x] `ruff check .` — clean

The tests pin the contract rather than the implementation: the zone name reaches the design dict, it is resolvable when present, the rendered `now_hhmm` describes the same moment as the zone it reports (so the label and the corrected marker cannot contradict each other), and the rendered page actually ships the zone to the client.

## Note

I have not touched `_build_timeline`. Between this and #286 it is the second live bug the tracker attributes to it; if it is dead it wants deleting rather than maintaining, but that is your call rather than something to fold into a fix.

### DIFF
--- a/app/deck_routes.py
+++ b/app/deck_routes.py
@@ -538,7 +538,17 @@ def _design_cards(
                 "cards": unbound,
             }
         )
-    return {"cards": cards, "groups": groups, "now_hhmm": now_tz.strftime("%H:%M")}
+    # The IANA name, not just the rendered time: the rail's now-marker ticks
+    # client-side, and a browser in a different zone from the configured one
+    # would otherwise recompute the mark against its own clock and jump by the
+    # offset a minute after load (#165, same class as #143 / #164 / #170).
+    tz_name = getattr(tz, "key", "") or ""
+    return {
+        "cards": cards,
+        "groups": groups,
+        "now_hhmm": now_tz.strftime("%H:%M"),
+        "now_tz_name": tz_name,
+    }
 
 
 @bp.get("")
@@ -628,6 +638,7 @@ def index() -> str:
         cards=design["cards"],
         groups=design["groups"],
         now_hhmm=design["now_hhmm"],
+        now_tz_name=design.get("now_tz_name", ""),
         # -- schedules forms ----------------------------------------------
         schedules=schedules,
         status=schedule_status,
--- a/templates/decks.html
+++ b/templates/decks.html
@@ -191,14 +191,33 @@
         .forEach((wash) => wash.classList.add('do-push'));
     }
     // Tick the now marks each minute so the rails stay honest.
-    const tick = () => {
+    //
+    // In the configured timezone, not the browser's. The server anchors
+    // now_pct to settings.app.timezone; a browser somewhere else recomputing
+    // from its own clock made the marker jump by the offset a minute after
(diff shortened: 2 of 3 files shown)


dmellok:
Merged, and that's the last of the batch. Six PRs in a morning, each with a real root cause, a test that fails on main, and an honest note where the issue tracker had the location wrong. That's a lot of tidy work in the parts of the app that don't get much attention, and it's appreciated.

Two things on this one: it needed a rebase over the "Unavailable displays" group that landed on main yesterday, which I resolved by keeping both hunks, and the page test now completes first-run setup so it asserts against the rendered rail rather than a redirect.

All six shipped as v0.390.0 on edge. If you want to keep going, the Discord is where most of the hardware testers and widget authors hang out, and it's easier to talk through something like the `_build_timeline` cleanup or the LAB preset question there than across issue threads: https://discord.gg/6qmwkGhGR7

Issue #259, opened 2026-08-24
Secrets: decide whether to persist a generated encryption key in the data folder


dmellok:
Secrets are encrypted at rest with a key that comes from `TESSERAE_SECRET_KEY`, or, when that isn't set, from an HKDF over `settings.app.session_secret_secret` ([secret_box.py](https://github.com/dmellok/tesserae/blob/main/app/secret_box.py)). The fallback exists so a fresh clone works without ceremony, and it survives ordinary restarts.

It does not survive a recreated container or a restored data folder with a regenerated session secret. When it doesn't, every stored secret becomes undecryptable and reads back as an empty string. Non-secret settings are untouched, so the install looks configured while the credentials are silently gone: the Home Assistant URL is still right and every `ha_sensor` reports the install as unconfigured.

v0.364.1 makes that state legible (the settings field is shown empty and marked for repair, and the widget error names the real cause), but the trap is still there. Anyone who never set `TESSERAE_SECRET_KEY` is one container recreation away from re-entering every credential they have.

Options, roughly in order of how much they help:

1. **Generate and persist a key on first run**, written into the data folder next to `settings.json`. Then the key travels with the data it protects, and the failure only happens if someone restores settings without the key file. Needs care: a key sitting beside the ciphertext is weaker than one in the environment, so this is a durability-versus-secrecy trade and probably wants to stay opt-out for people who do pin an env key.
2. **Warn at startup** when running on the derived fallback key, with the exact line to add to the compose file. Cheap, and turns a silent trap into a visible one.
3. **Docs**: make pinning the key part of the Docker install page rather than something you discover afterwards.

At minimum 2 and 3. 1 deserves a decision on the trade-off before anyone writes it.

Reported in the wild: a user upgraded, kept their dashboards and their HA URL, and lost only the token.

dmellok:
Options 2 and 3 shipped: a startup warning with the exact env line, and the Docker install docs pin the key before any credentials are stored. Option 1 is the open decision. Retitled to that.

PR #288, opened 2026-09-06
feat(secrets): warn while running on the derived encryption key (#259)

vaibhav8a:
## What this changes

Tesserae warns at startup while it is encrypting secrets with the derived fallback key, and the warning carries the exact line to fix it.

Part of #259 — items **2** and **3**, which the issue calls the minimum.

## Why

`resolve()` already noted the fallback, at `info`, which most deployments never render. The failure it is warning about is not a small one: the derived key survives an ordinary restart and **does not** survive a recreated container or a data folder restored alongside a regenerated session secret. When it does not, every stored secret decrypts to an empty string while every non-secret setting beside it is intact — so the install looks configured and is not, and nothing says so until the next time something tries to use a credential.

What comes out now:

```
WARNING TESSERAE_SECRET_KEY is not set, so secrets are encrypted with a key derived from
the session secret. That key does not survive a recreated container or a restored data
folder, and stored secrets then read back empty while the rest of the settings stay
intact. Pin one now — add to your compose environment:
    TESSERAE_SECRET_KEY=280e6c015a205aff1a7e277370e0c7666a7cf2d8feb00be5a8061e165e22731d
```

The issue asks for "the exact line to add to the compose file", so the warning generates one rather than telling the operator to go and find out how. The suggestion is fresh per call and not derived from anything — a derived suggestion would reproduce the coupling the warning exists to break.

`resolve()` runs once per boot from `app_factory`, so this is a startup warning by construction rather than by adding a second hook.

## Docs (item 3)

The Docker page gains **Pin the secret key before you store any credentials**, placed immediately after first-run and before anything that would store one — the advice is only useful before the secrets go in. It covers generating a key, where it goes, that adding it later protects everything entered after, and that the key does not belong in the same backup as the data it decrypts.

The configuration table described the derived key as one that "works fine; pin it explicitly if you rotate or template `settings.json`". That understated it, and is the sentence an operator would have read before losing a token. It now names the failure and links the new section.

## Not in this PR

Item 1 — generating and persisting a key beside `settings.json` — is left alone. The issue says it "deserves a decision on the trade-off before anyone writes it", and a key sitting next to the ciphertext it protects is exactly the kind of call that should be yours.

## Test plan

- [x] `pytest tests/test_secret_box.py tests/test_secret_key_warning.py -q` — 18 passed
- [x] `pytest -q -k "secret or settings or config"` — 226 passed
- [x] `ruff check . && ruff format --check .` — clean

The tests pin the properties rather than the wording: it warns at `>= WARNING`, the message contains a `TESSERAE_SECRET_KEY=<64 hex>` line, it names the *failure* and not just the setting (an operator is deciding whether to act now or later), a pinned key does **not** warn so nobody is nagged after acting, and the suggested key actually round-trips through `from_env` when pasted.

### DIFF
--- a/app/secret_box.py
+++ b/app/secret_box.py
@@ -70,6 +70,16 @@ class SecretBoxError(ValueError):
     so the user can fix it instead of getting empty secrets)."""
 
 
+def suggested_env_key() -> str:
+    """A fresh key in the exact form :data:`ENV_KEY` expects.
+
+    Generated per call rather than derived from anything: an operator pasting
+    this is choosing a new key, and one derived from the session secret would
+    reproduce the very coupling the warning is telling them to break.
+    """
+    return secrets.token_hex(KEY_BYTES)
+
+
 class SecretBox:
     """Holds a 32-byte AES-GCM key plus the wrap / unwrap helpers.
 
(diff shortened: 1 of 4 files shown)


dmellok:
Merged, thanks. Two adjustments on top, both about who reads the warning: the wording now covers every install type rather than assuming a compose file, and under the HA App it logs at info, since that install has no environment block to put the key in. I also tightened the docs to say the key is lost when the session secret is regenerated, rather than on any container recreate, because a compose recreate with the data volume keeps it.

Issue #251, opened 2026-08-22
Add CSRF tokens to session-authenticated form posts


dmellok:
Found during an audit of externally-contributed code. Not urgent, but worth closing.

## What is missing

There is no CSRF protection anywhere in the app: no `flask_wtf`, no `CSRFProtect`, no tokens in any form. Every state-changing admin POST (device delete, page edits, settings writes, plugin admin actions) accepts a cross-site form submission that carries the session cookie.

`SESSION_COOKIE_SAMESITE` is also not configured, so the default applies.

## Why the real risk is low

Browsers default session cookies to `SameSite=Lax`, which blocks cross-site POSTs. So the practical exposure today is narrow: an authenticated operator with a hostile page open, on a browser that does not apply the Lax default.

The loopback variant of this was worse and is already fixed (v0.329.0): plugin blueprint routes answered unauthenticated requests from loopback, and the server-side screenshot renderer can be pointed at `127.0.0.1`, so a rendered page could POST to the admin API with no credential at all. That path is closed; this issue is the remaining browser-side case.

## Suggested fix

1. Set `SESSION_COOKIE_SAMESITE = "Lax"` (or `"Strict"`) and `SESSION_COOKIE_HTTPONLY = True` explicitly rather than relying on browser defaults.
2. Decide whether tokens are wanted on top. Worth checking first whether any legitimate cross-site POST exists (an HA ingress flow, a bookmarklet, the Companion app) before turning them on, since that is what usually breaks.

dmellok:
SESSION_COOKIE_SAMESITE=Lax and SESSION_COOKIE_HTTPONLY are now set explicitly. CSRF tokens are still absent. Retitled to that.

PR #289, opened 2026-09-06
feat(auth): pin the session cookie's SameSite and HttpOnly (#251)

vaibhav8a:
## What this changes

`SESSION_COOKIE_SAMESITE = "Lax"` and `SESSION_COOKIE_HTTPONLY = True`, set explicitly instead of inherited from the browser.

Step 1 of #251. Step 2 (tokens) is not here — but the check the issue asks for first is done, below.

## Why

Every state-changing admin POST rides this cookie and the app carries no CSRF tokens, so what stops a cross-site submission today is the browser's own SameSite default. That makes the protection whatever the operator's browser decided, which is not a property the app can state.

Three decisions worth naming, since each is a place this could have gone wrong:

**Assigned, not `setdefault`.** My first version used `setdefault` and it silently did nothing: Flask ships `SESSION_COOKIE_SAMESITE` already present and set to `None`, so the attribute stayed off the cookie while the code looked like it set it. That is the same inherited-default failure the change exists to close, which is why the tests read the rendered `Set-Cookie` header and not only the config dict.

**`Lax`, not `Strict`.** Strict also withholds the cookie on a top-level GET arriving from another origin, so following a link to the dashboard from Home Assistant or a chat message would land the operator on a login page. Lax blocks the cross-site POST, which is the shape that matters here.

**`Secure` deliberately left off**, and pinned off by a test so nobody hardens it by reflex. These installs are overwhelmingly plain HTTP on a LAN or a Pi; a Secure cookie is not sent over HTTP at all, so setting it would log every one of them out rather than protect anything. An operator terminating TLS in front can set it after `create_app` returns.

## The check step 2 asks for

The issue says to check "whether any legitimate cross-site POST exists (an HA ingress flow, a bookmarklet, the Companion app) before turning [tokens] on, since that is what usually breaks". I looked:

- **Companion app** — authenticates with a scoped bearer token, and `companion_api.py` says so directly: *"a bearer token, so the token is the security boundary, not the origin"*. It does not ride the session cookie, so SameSite does not touch it and CSRF tokens would not either.
- **CORS surfaces** — `Access-Control-Allow-Origin: *` appears on three routes in `app_factory.py` (`/renders/`, `/page-assets/`, and one alongside). All are `@app.get`, image/asset reads for the device emulator's canvas. No POST.
- No `flask_wtf`, no other declared cross-origin POST consumer.

So nothing legitimate depends on a cross-site POST with the session cookie today. That does not decide step 2 — tokens are still a real cost to maintain on every form — but the blocker the issue was worried about does not appear to exist.

## Test plan

- [x] `pytest tests/test_session_cookie_attributes.py -q` — 4 passed; the SameSite assertion fails on `main` (`assert None == 'Lax'`)
- [x] `pytest -q -k "auth or session or login or cookie"` — 115 passed, 1 skipped
- [x] `ruff check .` — clean

### DIFF
--- a/app/app_factory.py
+++ b/app/app_factory.py
@@ -451,6 +451,31 @@ def create_app(
         apply_ha_options(settings)
     app.secret_key = auth.secret_key(settings)
 
+    # Session cookie attributes, set rather than inherited (#251). Every
+    # state-changing admin POST rides this cookie, and the app carries no CSRF
+    # tokens, so what stops a cross-site submission today is the browser's own
+    # SameSite default. Relying on a default means the protection is whatever
+    # the operator's browser decided, which is not a property this app can
+    # state.
+    #
+    # ``Lax`` and not ``Strict``: Strict also withholds the cookie on a
+    # top-level GET arriving from another origin, so following a link to the
+    # dashboard from Home Assistant or a chat message would land the operator
+    # on a login page. Lax blocks the cross-site POST, which is the shape that
+    # matters here.
+    #
+    # ``Secure`` is deliberately NOT set. These installs are overwhelmingly
+    # plain HTTP on a LAN or a Pi, and a Secure cookie is simply not sent over
+    # HTTP -- pinning it would log every one of them out rather than protect
+    # anything. An operator terminating TLS in front can set it themselves.
+    # Assigned, not ``setdefault``: Flask ships ``SESSION_COOKIE_SAMESITE``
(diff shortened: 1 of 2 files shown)


dmellok:
Merged, thanks. The Lax-not-Strict reasoning and the note on the setdefault trap were both useful to have written down. One follow-up on top: the header test now completes first-run setup so it inspects a real Set-Cookie instead of skipping on the 302 the fixture returns.

PR #290, opened 2026-09-06
feat(companion): support configurable personal-data retention

charmmmz:
## What this changes

Companion publishers can retain their latest Reminders, Reminders Fridge, and Health Summary snapshots for up to 365 days, or explicitly use `expires_at: null` to keep them until replaced or deleted. Servers advertise `personal_data_retention`; the deadline field remains required, and existing finite deadlines continue to work.

## Why

The current 48-hour maximum can remove a useful reminder or health summary between phone syncs. This supports the [Companion app's per-server retention controls](https://github.com/charmmmz/tesserae-companion-ios/commit/808f9ef) while keeping freshness separate: data becomes stale after 24 hours, and finite expiry still removes raw values. Storage remains latest-only and scoped to each paired publisher and source.

## Test plan

- [x] `.venv/bin/python -m pytest -n 6 -q` — 4,316 passed on Python 3.12.
- [x] `.venv/bin/ruff check . && .venv/bin/ruff format --check .`
- [x] `.venv/bin/python -m mypy app`
- [x] Regression coverage for 7/30/90/365-day and indefinite retention across all three sources, status responses, persistence after reload, explicit deletion, missing-deadline rejection, and changing Never back to finite expiry.
- Manual validation against a deployed server is not included in this PR's validation.

## Checklist

- [x] Tests and OpenAPI fixtures added for the new behavior.
- [x] Ruff and mypy clean.
- [x] Retention and widget behavior documented in `docs/integrations/companion-data-retention.md`.
- [x] User-facing entry added under `CHANGELOG.md`'s Unreleased section.
- No version bump or release tag; this PR only adds the API capability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Added
 
+- Companion publishers can choose snapshot retention up to 365 days or explicitly
+  retain the latest snapshot until replacement or deletion. The optional
+  `personal_data_retention` capability enables Never; freshness stays independent
+  of expiry, and existing clients retain their chosen finite deadlines.
+
 - Home Assistant MQTT discovery now exposes lineups and device
   operations. The hub device gains an *Automation* switch (pause every
   scheduled push), a *Quiet hours* switch, a switch per lineup with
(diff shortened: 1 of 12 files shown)


dmellok:
Merged, thanks. The null-guard coverage across the store, status listing and redaction sweep meant nothing else needed touching, and the reminders-fridge widget already handled a missing deadline. Two things on top: the retention page is now in the docs nav (it landed in a new directory that nothing linked to), and the server went to 0.391.0 on edge with the capability. The forward-compat CI failure was the flaky prewarm push test, not your change.

PR #291, opened 2026-09-06
feat(calendar_month): the "Week starts" label is optional, off by default (#283)

### LINKED ISSUE #283: [Feature]: Remove or make optional the "Week Starts {SUN,MON}" label in Month Widget
### What problem are you trying to solve?

The "Week Starts {SUN,MON}" label is silly and adds needless clutter, since the first column is already labeled with the day.

### Proposed solution

Remove the "Week Starts {SUN,MON}" label or make it optional.

### Alternatives considered

_No response_

### Additional context

_No response_

vaibhav8a:
## What this changes

The month widget's `Week starts Sun/Mon` header label is now a `show_week_start` option, off by default.

Closes #283

## Why

The reporter's point holds: the weekday column headers directly beneath already say which day the week starts on, so the label restated it and cost a line of header on a panel where space is the scarce thing.

**Made optional rather than removed, and defaulting off.** The widget has a `date_label_style` option with `minimal` (1–2 characters, `M Tu W Th…`) and `short` settings — with those the first column is much less self-evident, so there is a real case for the label. Just not the default one, which is what the issue is about.

## Test plan

- [x] `node plugins/calendar_month/tests/clamp_check.mjs` — passes
- [x] `pytest tests/test_widget_clamp_checks.py -q` — 4 passed
- [x] `pytest -q -k "calendar or clamp or manifest"` — 159 passed
- [x] Verified the new assertion actually fails: with `client.js` reverted, the self-check throws

The assertion goes in the widget's own `clamp_check.mjs`, which `test_widget_clamp_checks.py` already runs under pytest, so it is covered by the normal suite rather than needing a new harness.

It asserts against the source rather than a rendered DOM, because the label lives inside the render template literal and the property worth pinning is that it is *gated at all* — an ungated one reappears the moment someone edits that block.

## One note on the diff

My first commit rewrote `plugin.json` through `json.dumps`, which expanded every compact array in the file and turned a one-option addition into a 77-line diff. Re-done as a text insertion, so the manifest change is the 7 lines it should be and the file's existing formatting is untouched.

### DIFF
--- a/plugins/calendar_month/client.js
+++ b/plugins/calendar_month/client.js
@@ -143,6 +143,12 @@ export default function render(shadow, ctx) {
   const monthDate = new Date(data.year || 2026, (data.month || 1) - 1, 1);
   const monthName = localizedFull(monthDate, "month", locale).toUpperCase();
   const year = data.year || "";
+  // Off by default (#283): the weekday column headers already say which day
+  // the week starts on, so the label restated it and cost a line of header on
+  // a panel where space is the scarce thing. Kept as an option rather than
+  // deleted, because a minimal or single-letter weekday style makes it much
+  // less obvious at a glance.
+  const showWeekStart = opts.show_week_start === true;
   const weekStartLabel = (data.week_start === "sunday")
     ? t("week_start_sunday", "Week starts Sun")
     : t("week_start_monday", "Week starts Mon");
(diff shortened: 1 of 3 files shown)


dmellok:
Merged, thanks. Keeping it as an option for the minimal weekday style was the right call. Bumped the widget to 0.3.1 alongside.

PR #293, opened 2026-09-06
feat(calendar): use per-event colours from the .ics feed (#222)

### LINKED ISSUE #222: [Feature]: Use event colors as defined in .ics file
### What problem are you trying to solve?

My family calendar is just a single Google Calendar, but we extensively colorize events to categorize them (e.g. work vs. school vs. vacation). I'd like to be able to see all these colors on the eink display.

### Proposed solution

The Calendar Core component should pass along the custom event color from the ics feed to the widget so it can colorize based on those. It seems reasonable that this could be an optional thing as overriding the colors to set a single color per calendar could be really useful in other cases.

### Alternatives considered

I *could* add the same family calendar many times as a bunch of separate feeds and try to use filters on each feed somehow so that each "feed" could be colorized to match. But a) that's not very convenient, and b) I'm not sure I can filter on color, which is our only differentiator in the calendar.

### Additional context

_No response_

vaibhav8a:
## What this changes

An event's own colour, where the calendar sets one, can now drive the widget colour instead of the feed's single colour.

Closes #222

## Why

A single family calendar is often colour-coded by category — work vs school vs vacation — and every one of those events reached the widget stamped with `feed_colour`, so the distinction the user maintains upstream was thrown away on the way in. The reporter's workaround was adding the same calendar many times and filtering, which they note they cannot do because colour is the only differentiator.

## What it reads

Two spellings, because producers disagree:

- **RFC 7986 `COLOR`** — a CSS3 colour *name*, which is what Google Calendar writes.
- **`X-APPLE-CALENDAR-COLOR`** — a hex triple, sometimes with an alpha byte, which is what Apple and several CalDAV servers write.

An unrecognised value yields **no** colour rather than a guess, so the event falls back to the feed's. Inventing one would put a wrong colour on an event whose whole point is that the colour carries meaning.

## Opt-in per feed

As the issue suggests ("It seems reasonable that this could be an optional thing"), and per feed rather than globally, because the two uses are opposite: one calendar categorises by colour, another wants every event to read as "work" whatever the producer set.

A palette button on each feed row toggles it. The colour is carried on *every* event and only the precedence is gated, so flipping the toggle needs no cache drop — and an event with no colour of its own always falls back, so a partly-coloured calendar stays coherent rather than going half-default.

| opt-in | event colour | result |
|---|---|---|
| on | `#ff6347` | `#ff6347` |
| on | none | feed colour |
| off | `#ff6347` | feed colour |
| off | none | feed colour |

HA-sourced feeds do not get the toggle: their events come from the HA calendar API, not an ICS body, so there is no per-event colour to read.

## Test plan

- [x] `pytest plugins/calendar_core/tests/test_event_colours.py -q` — 8 passed
- [x] `pytest -q -k "calendar or feed or ics or colour"` across `tests/`, `calendar_core`, `calendar_day` — **147 passed**, 6 skipped
- [x] `ruff check .` and `ruff format --check .` — clean

**One thing I broke and fixed, worth flagging.** My first version of the test did `sys.path.insert` + `import server`, copying the sibling plugin suites. That turned **eleven passing `calendar_day` tests red** without touching them: every plugin ships a module called `server`, so whichever suite imports first owns `sys.modules["server"]` and the other gets the wrong module. This file now loads its own `server.py` under a unique name via `importlib`, and the docstring says why so the next person copying a sibling does not reintroduce it.

That collision is latent between the existing plugin suites too — it only bites when two of them run in the same session. Not something I have touched here beyond not adding to it.

### DIFF
--- a/plugins/calendar_core/server.py
+++ b/plugins/calendar_core/server.py
@@ -442,6 +442,11 @@ def _expand_events_full(blob: bytes, start: datetime, end: datetime) -> list[dic
                     "start": s_iso,
                     "end": e_iso,
                     "all_day": all_day,
+                    # The event's own colour, when the producer set one
+                    # (#222). Carried on every event and honoured only when
+                    # the feed opts in, so turning the option on and off does
+                    # not need a cache drop.
+                    "event_colour": _event_colour(ev),
                 }
             )
         except Exception:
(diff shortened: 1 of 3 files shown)


dmellok:
Merged, thanks. I pushed one commit onto your branch before merging: the comments said Google Calendar writes `COLOR` and that Apple writes `X-APPLE-CALENDAR-COLOR` per event. Google's export carries no colour at all, which is what the thread on #222 established, and Apple's property is calendar-level. The code is unchanged and still accepts a hex value found on an event, since feed bridges do put it there. In 0.393.0.

Issue #274, opened 2026-09-01
[Bug]: Ghosting on webhook triggered update, but not on normal updates


twpeltenburg:
### Tesserae version

0.379.0

### How is Tesserae running?

Native (install.sh / install.ps1)

### Operating system

Debian 12

### Panel hardware

ReTerminal E1003

### Client repo + version

v1.27.2

### Steps to reproduce

I have a dashboard that show buttons, a button press triggers a webhook to a server. That server sends a webhook back to tesserae to update and push the dashboard. See below for the buttons layout.

<img width="2048" height="1536" alt="Image" src="https://github.com/user-attachments/assets/2662ebd8-c223-4a99-a2bf-b86bbcd99634" />

### Expected behaviour

I expect the display to refresh and show the new content. An example:

<img width="2048" height="1536" alt="Image" src="https://github.com/user-attachments/assets/9186052b-921e-4560-a04a-eadee16e844c" />

### Actual behaviour

When I push the button, the webhooks get sent, the panel refreshes but leaves a ghost image. This does not happen on normal (timed wakeup) refreshes. Maybe it happens because it is a high contrast area, but it doesn't happen in normal refreshes, so that's weird.

<img width="2048" height="1536" alt="Image" src="https://github.com/user-attachments/assets/6a55a04a-2083-4d07-9105-687fee852d25" />

### Server logs

```shell

```

### Before submitting

- [x] I searched [existing issues](https://github.com/dmellok/tesserae/issues?q=is%3Aissue) and didn't find a duplicate.
- [ ] If I could, I tried reproducing on the latest minor release.

dmellok:
Timed wakeups paint the full frame at full quality, but a webhook push sends only the changed regions, and the firmware refreshed those with the fast waveform used for touch feedback, which ghosts on high-contrast areas.

Fixed in v1.27.3: pushed updates now use the same full-quality refresh as a scheduled repaint, at a cost of about 0.2 s per update. Your panel will get it over the air shortly, or flash from https://tesserae.ink/flash. Feel free to close this once it's cleared up on your setup.

twpeltenburg:
Unfortunately the behavior is still the same. See the video attached of the refresh sequence, you can see exactly what happens if you play it frame by frame. 

https://github.com/user-attachments/assets/2d4cac54-a251-4cfc-95c1-ba6cafce0058

dmellok:
[...]

PR #294, opened 2026-09-06
docs(devices): explain ghosting after a pushed update (#274)

vaibhav8a:
## What this changes

A **Ghosting after a pushed update** section under *Per-device settings*, connecting the symptom to the control that fixes it.

Refs #274. Docs only — no behaviour change.

## Why, and what I found

The reporter's observation is the diagnostic: *"This does not happen on normal (timed wakeup) refreshes."*

**Update delivery** defaults to *Auto*, which sends a small change as a partial-refresh patch (`_divert_to_patches_locked` in `app/push.py`). A partial refresh does not run the panel's clearing waveform, so a high-contrast change leaves the old content faintly behind. A device that was asleep cannot apply a patch at all, so its next poll is served a whole frame and a full repaint — which is exactly why the timed wake looks right and the button-triggered push does not.

`refresh_mode: "full"` disables the patch divert per device, and the control already says *"use it if updates look incomplete or ghosted"*.

**So the remedy exists and the control is honest.** What was missing is anything connecting the symptom to that control for someone who has not already found it: a search of `docs/` for "ghost" turns up only widget-design guidance about avoiding pure `#000`/`#fff`, nothing about delivery mode. Someone hitting this reaches for the tracker, which is what happened.

## What this does not do

It does not stop the ghosting. Whether *Auto* should decline to patch a high-contrast or large-area change on its own is a real question — the reporter's buttons are exactly that shape — but it is a heuristic with a cost on every push and a call for you rather than something to fold into a docs fix.

I have not closed the issue in the commit for that reason; if you would rather it stay open for the heuristic, this just makes the workaround findable in the meantime.

## Test plan

- [x] `pytest -q -k "docs or markdown or mkdocs or nav"` — 32 passed

### DIFF
--- a/docs/install/devices.md
+++ b/docs/install/devices.md
@@ -44,6 +44,26 @@ renderer, so two panels driven by the same renderer can differ. For TRMNL /
 1-bit panels the dither dropdown carries the full set: Floyd-Steinberg,
 Atkinson, Jarvis-Judice-Ninke, Stucki, Bayer 8×8, halftone, crosshatch, or none.
 
+### Ghosting after a pushed update
+
+A panel that repaints cleanly on its timed wake but leaves a faint ghost of
+the previous frame after a push — a webhook, a manual Send, a button action —
+is showing you the difference between the two delivery paths, not a fault.
+
+**Update delivery** on the device card defaults to *Auto*, which sends a small
+change as a partial-refresh patch on firmware that supports it. A partial
+refresh does not run the panel's full clearing waveform, so a high-contrast
+change can leave the old content faintly behind. A device that was asleep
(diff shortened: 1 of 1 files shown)


dmellok:
Merged with a rewrite, so worth explaining. The thread on #274 ended somewhere different from where this section started: frame by frame, that panel was painting full frames, so the two patch-path fixes did nothing for it and the cause traced to the panel controller's reference image after a touch wake, handled in firmware 1.28.1. Pointing that reporter at Always full refresh would not have helped.

The gap you found is real though. The device guide never mentioned Update delivery, so the section now describes the control itself, when to reach for full refresh, and that ghosting on timed refreshes or after a touch is a firmware matter. Same place in the guide, no #274 link.

PR #295, opened 2026-09-06
feat(mcp): appearance is its own read, not part of every catalog read (#257)

vaibhav8a:
## What this changes

`/api/mcp/catalog` reports how many themes, styles and fonts exist and points at a new `GET /api/mcp/appearance` that serves them, instead of inlining the lists in every read.

Option **3** from #257. Option 1 (summarised `desc`, full text on `get_widget_options`) already landed.

## Why

The lists are only needed when something is actually being restyled, which most builds never do — and they were 18% of the single most expensive thing an MCP agent does. This is the same shape the icon set already uses in this endpoint (`total`, `search_endpoint`, `usage`), for the same reason, so it is a pattern the file already establishes rather than a new convention.

## Measured

Against the bundled widget set, `/api/mcp/catalog` before and after:

| | before | after | |
|---|---|---|---|
| **catalog total** | 18,433 B (~4,608 tok) | **15,247 B (~3,811 tok)** | **−17%** |
| `appearance` | 3,354 B (~838 tok) | 168 B (~42 tok) | −95% |
| `widgets` | 13,665 B | 13,665 B | unchanged |
| `libraries` | 1,113 B | 1,113 B | unchanged |

The issue measured the catalog at 22,765 B / 5,691 tokens. Option 1 took it to 18,433 B; this takes it to 15,247 B — together **a third off** the number in the issue.

`/api/mcp/appearance` serves the full 3,354 B when asked.

## Nothing lost, only moved

The issue's warning is the right one — *"if a one-line summary makes an agent pick worse widgets, the saving isn't real"* — and the equivalent risk here is an agent that can no longer find the themes. So:

- the catalog carries a **count** of each list, which is what tells an agent the endpoint is worth calling at all
- a test asserts those counts **equal the lengths** of the lists the endpoint serves, so the summary cannot drift into a lie
- a test asserts the summary stays a *pointer* — it fails if the appearance block grows back past a fifth of what the endpoint serves
- the catalog docstring now says where the lists went, since that docstring is what the agent reads

The new route sits behind the same experiment gate as the rest of the surface, pinned by a test — a new endpoint should not become a way around it.

## Test plan

- [x] `pytest tests/test_mcp_appearance_split.py -q` — 6 passed
- [x] `pytest tests/test_mcp_api.py tests/test_mcp_appearance_split.py -q` — 105 passed
- [x] `pytest -q -k "mcp or catalog"` — 171 passed
- [x] `ruff check . && ruff format --check .` — clean

### DIFF
--- a/app/mcp_api.py
+++ b/app/mcp_api.py
@@ -387,9 +387,11 @@ def _desc_summary(desc: str) -> str:
 
 @bp.get("/catalog")
 def catalog() -> Response:
-    """Every renderable widget (with its fragments) plus theme/style/font options,
-    the vendored code-element libraries, and the icon set, so the agent knows what
-    it can place, how to style the canvas, and what's in the code-element toolkit.
+    """Every renderable widget (with its fragments), the vendored code-element
+    libraries, and pointers to the icon set and the appearance options, so the
+    agent knows what it can place and what's in the code-element toolkit.
+    Theme / style / font lists are counted here and served in full by
+    ``GET /appearance``; the icon names by ``GET /icons?q=``.
 
     The per-widget ``sample`` payload is omitted here to keep the response small;
     fetch a widget's live data shape with ``POST /widgets/<key>/data`` instead.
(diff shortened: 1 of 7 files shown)


dmellok:
Merged, with one commit on top that the change needed. The server side was right, but the bridge is the consumer, and it passes the appearance block through untouched: after this PR, `list_widgets()` hands an agent `appearance: {themes: 4, ...}` while the canvas-document shape and the list_widgets tool doc still said to take theme, style and font ids from `list_widgets().appearance`. Those docs live in `app/mcp_docs.py` and are served to the bridge at handshake, not the Flask route docstring. No bridge tool reached the new endpoint, so an agent following the docs had no way to get a font name.

The added commit routes `list_widgets(section="appearance")` to `GET /appearance`, updates the served docs and the bridge's embedded copies, and pins both with tests. That is a bridge behaviour change, so it went out as tesserae-mcp 0.16.0 alongside server 0.393.0. Worth knowing for next time: the bridge tests mock the catalog, so a catalog shape change will not fail them.

PR #303, opened 2026-09-10
feat(hardware): add the original M5Stack M5Paper (IT8951)

Zrad:
Adds a hardware-catalog entry for the original **M5Stack M5Paper** (pre-S3),
the 4.7" 960x540 16-level greyscale handheld on the classic ESP32-D0WDQ6-V3.

Companion to dmellok/tesserae-device-firmware#33, which adds the
`m5stack-m5paper` build. **Verified end-to-end on hardware 2026-09-10**: dashboard frame,
battery telemetry, the three side buttons, the SHT30 sensor, and touch.

Catalog-only on this side. Unlike the PaperS3, the M5Paper reaches its glass
through an IT8951 controller over SPI, so the firmware reuses the reTerminal
E1003's greyscale driver -- but on the wire it is the PaperS3's geometry, and
`esp32_gray_bin` already produces exactly the 259200-byte frame it wants. No
new renderer, no new packer, no server code.

### Why `auto_select: false`

An M5Paper and a PaperS3 report the same relay model (`esp32_client`), the same
gamut (`gray_16`) and the same 960x540 scan at pairing. Nothing on the relay
wire distinguishes them, so inference would be a coin flip. This kind is only
ever set explicitly -- pinned by the operator, or declared by the firmware's
`TESSERAE_DEVICE_KIND`. An ambiguous report keeps resolving to
`m5stack_papers3` exactly as it does today, which
`tests/test_relay_home.py` covers.

### Touch

`touch: true`, with the usual `touch_enabled` / `touch_linger_s` form and no
buzzer fields (this board has no piezo). The GT911 reports in the 540x960
composition space, so taps hit-test directly against the dashboard's regions
with no server-side mapping. It needs Stay awake on: the classic ESP32 cannot
wake from deep sleep on the touch INT -- the same ext1 limitation that caps the
side buttons -- so a tap only registers while the panel is awake and polling.
The help text on the field says this.

### Docs

`docs/compatibility.md` is regenerated, not hand-edited -- the prose lives in
`scripts/gen_compatibility.py`'s vendor intro and in `docs/_data/tested.json`,
both updated here. The `esp32_gray_bin` row now separates what is confirmed
(E1003, and now the M5Paper) from what still awaits a bring-up (EE03, PaperS3).

Full test suite passes (4015).


dmellok:
Thanks, merged. The one ruff complaint (quote style on line 96 of scripts/gen_compatibility.py) I fixed on main in 8a4d833 along with the bump to 0.406.1. Full suite passes on the branch locally, and the regenerated compatibility page matched the committed one byte for byte.

PR #306, opened 2026-09-11
fix(rest): clear the scheduler tick when waking a device for a lineup

partridgeworks:
## What this changes

`next_poll_s` for a scheduler-projected change now carries a 70 s margin instead of 20 s, and a change overdue by less than a scheduler tick is treated as imminent ("poll again after the margin") instead of being skipped as stale. The 20 s widget-hint margin is unchanged, since `/frame` already re-renders on demand for those.

## Why

A sleeping REST device woken for a daily lineup collected the *previous* day's frame most mornings.

- `/status` told the device to come back 20 s after the lineup's target. But the scheduler fires on a 30 s tick whose phase is set by process start, then renders each due lineup in turn, so the new frame landed anywhere up to ~40 s after the target. The poll arrived before it existed.
- The miss then stuck. On that early poll the projection correctly reported the lineup as "due now" (the next tick fires it), but `scheduled_at` is truncated to whole seconds, so it reached `_projected_poll_s` a fraction of a second in the past, tripped `if delta < 0: continue`, and the device was told to sleep its whole configured interval on the old frame.

Observed on an Inkplate 10 (`esp32_bw_client`, 6 h `sleep_interval_s`, daily lineup at 06:00 Europe/London) over four mornings, from the battery history (every heartbeat) and the event log (every fire):

| Day | Device polled | Lineup rendered | Result |
|---|---|---|---|
| Mon | 06:00:22 | 06:00:24 | miss, fixed by button press |
| Tue | 06:00:26 | 06:00:14 | hit |
| Wed | 06:00:21 | 06:00:36 | miss, fixed by button press |
| Thu | 06:00:22 | 06:00:37 | miss, fixed by button press |

Each morning the device first woke at ~05:58 (its 6 h grid pulled forward), was told `next_poll_s` ≈ 110 (target + 20 s), came back at 06:00:2x, and then slept 21 600 s.

## Test plan

- [x] `pytest -q tests/test_rest_api.py tests/test_rest_always_on.py tests/test_wake_alignment.py tests/test_device_upcoming.py`
- [x] `mypy app`
- [x] New tests: an event 0.5 s overdue yields the margin rather than the configured interval; an event 120 s overdue is still ignored. Existing margin-pinned assertions updated from 20 to 70; the always-on floor test now exercises the widget-hint path, which is where the 20 s case still lives.
- [ ] Manually verified: deployed to the home server above. The real check is the Inkplate's next 06:00 wake landing on the fresh frame without a button press; not yet observed at the time of opening this PR.

Not covered by automated tests: the end-to-end timing against a real scheduler thread, which is what the hardware check above is for.

## Checklist

- [x] Tests added for new behaviour
- [x] `ruff` + `mypy` clean
- [x] User-facing changes documented (`docs/dev/client-protocol.md`, projection section)
- [x] `pyproject.toml` version bumped (0.408.2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Fixed
 
+- A sleeping REST device woken for a daily lineup no longer collects the
+  previous day's frame. The device was told to poll 20 s after the lineup's
+  target, but the scheduler fires on a 30 s tick and renders each due lineup
+  in turn, so most mornings the poll landed before the new frame existed. The
+  miss then stuck: the projection reported the lineup as due now, a
+  fraction of a second in the past once truncated to the second, and the
+  poll path skipped it as stale and sent the device to sleep for its whole
+  interval on the old frame. The scheduler margin is now 70 s, and a change
+  overdue by less than a tick means "poll again after the margin".
+
 - The Location picker on the weather, sky, and sunrise widgets now accepts a
   pasted `lat, lon` pair. It only ever asked the city name search, which
   returned "No matches." for coordinates, so a spot with no nearby town could
--- a/app/rest_api.py
+++ b/app/rest_api.py
@@ -531,13 +531,24 @@ def _configured_poll_s(device: Device) -> int:
     return 60
 
 
-# Seconds to add to a projected content change before telling the client to
-# poll. The scheduler fires at the projected instant and the render itself is
-# a browser compose plus a quantize, so a client polling at exactly that
-# moment races the render and collects the *previous* frame. A margin costs
-# nothing (the device is asleep for it) and turns a guaranteed miss into a
-# hit.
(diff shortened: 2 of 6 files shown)


dmellok:
Both in on 0.410.1, this and #308. The timing table in the description made this an easy one to take: the projection does report an overdue-but-unfired record as due now with the microseconds dropped, so the sub-second skip was exactly where you said it was.

Two edits on the way in. I widened the new poll-timing assertions by a few seconds, since a loaded CI runner has tripped 1 to 2 s windows before, and I added a CHANGELOG line for the dither clamp.

Did the Inkplate's next 06:00 wake land on the fresh frame? That's the one thing the tests can't cover, and I'd like to note it as confirmed on hardware before the next tag.

partridgeworks:
Hello, just to close this one off — I can confirm that the Inkplate’s next refresh landed on the correct frame so the issue seems resolved by this.

Issue #279, opened 2026-09-04
[Feature]: Add new translations for widgets


Frankynov:
### What problem are you trying to solve?

So for now I think there are mainly English, and I saw a bit of French, options for the widgets.
I was wondering what would it take to be able to give extra translations for existing official widgets, like slovak for example ?

### Proposed solution

Either put a configurator for strings to be editable by the user, or embed the translations directly in widget package - that means the translations would be in the source of the widgets themsleves

dmellok:
The second option is how it already works. A widget declares `"locales": [...]` in its plugin.json and ships a flat `strings/<tag>.json` per language next to it; client.js reads them through `ctx.t(key, fallback)`. The contract and the file shape are in the "Locales & strings" section of [docs/widgets.md](https://github.com/dmellok/tesserae/blob/main/docs/widgets.md#locales--strings).

Two caveats. Only the three calendar widgets (calendar_day, calendar_week, calendar_month) have opted in so far, each with English and French, so a Slovak PR today would only reach those. And the language picker in Settings is currently a fixed English/French list, so `sk` can't be selected until that's widened. I'll change the picker to list whatever languages the installed widgets ship, and start moving the other bundled widgets' hardcoded labels behind `ctx.t()` so there's something to translate.

If you want to get going now, this is the whole contribution. The format is a flat JSON object, keys identical to `strings/en.json`, values plain strings, no nesting or placeholders. The Slovak below is my rough attempt so you can see the shape; please correct it.

`plugins/calendar_day/strings/sk.json`
```json
{
  "no_events": "Dnes žiadne udalosti.",
  "event": "udalosť",
  "events": "udalosti",
  "at": "o"
}
```

`plugins/calendar_week/strings/sk.json`
```json
{
  "this_week": "TENTO TÝŽDEŇ",
  "event": "udalosť",
  "events": "udalosti"
}
```

`plugins/calendar_month/strings/sk.json`
```json
{
  "week_start_sunday": "Týždeň začína ne",
  "week_start_monday": "Týždeň začína po"
}
```

And in each of the three `plugin.json` files:
```json
"locales": ["en", "fr", "sk"]
```

[...]

PR #311, opened 2026-09-11
feat(i18n): weather_now_scenic widget french translation

theiereman:
## What this changes

Translates the weather_now_scenic widget to French

## Why

Make the widget accessible to french speaking users

refs #279

## Test plan

- [x] `pytest -q`
- [x] `mypy app`
- [x] Manually verified: translations are accurate

## Checklist

- [x] Tests added for new behaviour (or note why none made sense)
- [x] `ruff` + `mypy` clean
- [x] CHANGELOG entry added under `## [Unreleased]` if user-facing
- [x] `pyproject.toml` version bumped if this is going to be tagged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ All notable changes to Tesserae are recorded here. Format loosely follows
 
 ### Added
 
+- The weather_now_scenic widget's panel text (the current-conditions label
+  and the error-state title) is now available in French, following the same
+  `locales` / `ctx.t()` contract as weather_now and the calendar_* widget
+  family (docs/widgets.md#locales-strings).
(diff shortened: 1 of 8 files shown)


dmellok:
All five landed in 0.410.0: #305, #307, #309, #310 and this one. The four weather widgets are the first bundled widgets outside the calendar family to ship a second language, and the contract tests pinning the client-side mirror tables against server.py are the part I'd have got round to last, so that's a real help.

Two small edits on the way in. The four CHANGELOG bullets became one, and the hourly NOW chip reads "MAINTENANT" rather than "ACTUEL" so it matches the "Maintenant" legend entry underneath it. Shout if you'd have picked differently.

PR #313, opened 2026-09-14
feat(trmnl): 16-level greyscale renderer + device-card renderer picker

rpcai:
## What

Adds `trmnl_png_gray16`, a 16-level greyscale renderer for the TRMNL BYOS
transport, plus a renderer picker on the device card so it can actually be
selected.

## Why

TRMNL BYOS only shipped `trmnl_png` (1-bit B/W). Panels that paint real greys
over the same `/api/display` path lose most of that capability to 1-bit output:

- **Seeed reTerminal E1003 / TRMNL X** — the catalog already describes the E1003
  as "10.3 inch monochrome ePaper terminal **with 16-level greyscale** … speaks
  Tesserae's TRMNL BYOS endpoints once flashed with TRMNL firmware", but on that
  path it gets 1-bit.
- **Jailbroken e-readers running KOReader** (Kobo Clara HD, Kindle Paperwhite) —
  8-bit greyscale framebuffer, `trmnl.koplugin` hands the PNG straight to
  `RenderImage` which decodes greyscale fine.

On a text+chart dashboard the visible cost of 1-bit is that every weather-chart
area fill and shaded forecast band turns into coarse error-diffusion speckle
(worse still when the client upscales a smaller composition). A weather band that
`trmnl_png` renders as stipple, `trmnl_png_gray16` renders as three flat greys.

## The renderer

`renderers/trmnl_png_gray16/` mirrors `trmnl_png`'s pipeline exactly — fit /
flip / underscan / pre-dither contrast / dither — and changes only the quantise
target and container:

- quantise against `quantizer.GRAY_16_PALETTE` (the same 16-entry ramp
  `esp32_gray_bin` uses) instead of `_MONO_PALETTE`
- save mode `L` (8-bit greyscale, ≤16 distinct values) instead of `convert("1")`
- honour a device's measured grey ramp when a greyscale calibration profile is
  applied, via the same `settings["_gray_ramp"]` side channel `esp32_gray_bin`
  reads, so error diffusion aims at greys the panel actually produces

Same 8 dither modes and `contrast` slider as `trmnl_png`, both
`device_setting: true`. 8-bit rather than 4-bit PNG on purpose: Pillow's 4-bit
support is patchy and several e-reader decoders handle `L` cleanly where 4-bit is
a coin toss; the data only spans 16 values so deflate keeps the file close to a
4-bit encode anyway.

`devices/trmnl_client/device.json` now lists both renderers, `trmnl_png` first so
every existing device is unchanged.

## The picker

`renderer_id_for_format` disambiguates renderers by file extension, and both
`trmnl_png` / `trmnl_png_gray16` are `.png`, so there was no way to select the
new one short of hand-editing the instance manifest. (`trmnl_png_color` has the
same latent problem.)

Added a **Renderer** `<select>` to the device card (Settings → Devices →
General), shown **only** when a kind exposes more than one renderer of the same
wire format that the extension-matching auto-select can't tell apart — today
that is `trmnl_client` (1-bit `trmnl_png` vs 16-grey `trmnl_png_gray16`).
Backed by `device_service.set_instance_renderer_id()`: repin on disk → reload
in place → `clone_for_instances` → drop the now-stale render (it may be a
different format) so the device re-renders on its next poll. `None` / the
kind's primary clears the pin so the instance follows the kind default again.

## Out of scope (follow-ups)

- Hardware catalog SKUs that select this renderer — a `seeed_reterminal_e1003_gray`
  variant (mirroring the existing `_gray` SKU convention) and a
  `kobo/clara_hd.json` entry. Kept separate since each wants its own on-hardware
  confirmation and a `docs/compatibility.md` line.

## Testing

- `renderers/trmnl_png_gray16/tests/test_smoke.py` — loader registration, BYOS
  wire contract, 8-bit / ≤16-level output, `dither="none"` snaps to the nominal
  ramp, `_gray_ramp` override honoured, self-contained payload URL.
- `tests/test_device_service.py::test_set_instance_renderer_id_pins_and_clears` —
  pin / no-op / clear / reject-unknown.
- `ruff check`, `ruff format --check`, `mypy app` clean.
- Full `pytest` suite (4,870 tests) run in 8 memory-bounded batches on the
  dev host (a full single-process run OOM's the box on unrelated,
  pre-existing memory pressure) — all green, 0 failures.
- Built the image and ran it against a live instance: the picker renders with
  both options, switching repins + drops the stale render + re-renders, and
  `/api/display` serves an 8-bit 16-level PNG. Confirmed KOReader on a Kobo
  Clara HD paints it correctly (dithering gone, smooth greys).

https://claude.ai/code/session_01AAx1nPPjEnCYkuSQPwvmnx


dmellok:
Merged as 0.413.0, thank you. The renderer and the picker are in as written; the only changes on the way in were settling the version against main and marking the E1003 / TRMNL X path as untested until someone confirms that stock TRMNL firmware paints an 8-bit grey PNG. The Kobo result stands as the confirmed target.

The two follow-ups you listed, an E1003 grey SKU and a Kobo Clara HD catalog entry, are noted and open.

Issue #299, opened 2026-09-07
[Feature]: Quiet days


twpeltenburg:
### What problem are you trying to solve?

I use Tesserae in an office environment, I don't care about the displays refreshing during weekends. In order to save battery, add a "Quiet days" option so I can have the displays sleep over the weekend when the office is closed.

### Proposed solution

Maybe add a quiet schedule? eg. monday-friday 20:00-08:00, sat sun full day. So it's customizable on a daily basis.

dmellok:
Built this as an extension of quiet hours rather than a separate schedule. The quiet-hours window, both the global one and a device's override, now has two weekday pickers: the days the start/end window applies on, and days that are quiet all day. Your setup is 20:00 to 08:00 on Monday to Friday plus Saturday and Sunday all day, which runs quiet from Friday evening through to Monday morning.

There's also a new "Sleep through quiet hours" switch, which is the part that saves battery: instead of waking on its interval inside the window and finding nothing, the panel is told to sleep until the window ends. It's off by default because a manual push made during the window then waits until the window ends.

Shipped in server 0.398.0 on the edge tag; no firmware change needed. Docs are in the Quiet hours section of the devices page.

twpeltenburg:
Great implementation this way, thanks.

Issue #250, opened 2026-08-22
refresh_floor_s is no longer enforced anywhere


dmellok:
A panel's `refresh_floor_s` describes how fast its glass can be repainted. Every hardware profile that declares one (Seeed, Xteink, TRMNL, Waveshare at 60; Pimoroni Inky, Waveshare 13.3 at 30) now states a constraint the server does not apply.

## How it got here

Its only enforcement point was `_configured_poll_s` in `app/rest_api.py`, which clamped an always-on device's **poll** cadence to it. That conflated two different things: how often the device asks, versus how fast the panel can repaint. A poll is a conditional GET that answers 304 while the frame is unchanged, and a 304 never reaches the glass. The clamp meant an E1003 (floor 60) set to a 5 second always-on cadence was told to poll every 60, so a manual Send took up to a minute to appear. Removing it (v0.332.0) fixed that and left the repaint side unguarded.

For always-on panels the slow poll had also been *incidentally* rate-limiting frame delivery. That is now gone too.

## Why it has not bitten yet

Content only changes when something renders, and nothing renders fast enough today to drive a panel past its floor. The guarantee is missing rather than actively violated.

## Where it belongs

The repaint limit belongs on the path that decides to send a new frame, not on the poll cadence. Sketch:

- Track the last paint time per device.
- When a newly rendered frame differs from what is on the glass and the floor has not elapsed, hold it and paint when the floor expires rather than dropping it.
- A newer frame arriving during the hold replaces the held one, so the panel always lands on the latest content.
- Log a held frame with the remaining wait: from the operator side a deferred paint and a missed Send look identical.

The device is arguably the better enforcement point, since it alone knows its panel, but the server should not silently carry a field it ignores.

Issue #243, opened 2026-08-21
Let a widget declare when its data next changes, and feed that into next_poll_s


dmellok:
Follow-on from #241, and a blocker for the `room_status` widget.

## Problem

`_next_poll_s` consults `project_upcoming`, which projects schedules and rotation steps. It cannot see inside widget data, so any widget whose displayed state changes on its own clock is invisible to it.

A meeting-room panel is the clearest case. Its state flips when a meeting starts or ends, which is neither a schedule nor a rotation step. Set the generous backstop interval that #241 otherwise makes correct for a battery panel, and the panel sleeps straight through 15:00 and keeps showing BUSY. The workaround is a short polling interval, which throws away the benefit of #241 and burns wakes on a panel that mostly has nothing to say.

It generalises well past room status: bin collection, alarms, sunrise and sunset, countdowns, anything with a known next transition.

## Proposed

- A widget's `fetch()` may return a `next_change_at` (ISO 8601) alongside its data, meaning "the rendered output becomes wrong at this instant".
- The composer records the soonest such value across the widgets bound to a device.
- `_next_poll_s` folds it in beside the schedule projection, subject to the same rules already in place: the configured interval stays the ceiling, and the render margin still applies.

## Notes

- Purely additive. A widget that returns nothing behaves exactly as now.
- Wants a sanity clamp so a widget cannot pull a device into a hot-poll loop; the existing floor in `_next_poll_s` covers most of it.
- Worth deciding whether the value survives in the render cache or is recomputed, since a stale `next_change_at` is worse than none.

Issue #241, opened 2026-08-20
Derive next_poll_s from projected content changes, not just sleep_interval_s


dmellok:
Raised in [discussion #190](https://github.com/dmellok/tesserae/discussions/190#discussioncomment-18078954).

## Problem

`_next_poll_s` in `app/rest_api.py` resolves the per-device configured `sleep_interval_s`, then the kind schema default, then a 60 s fallback. It carries no knowledge of when the device's content actually next changes.

The docs and the field name both imply "new content is probably available at", which is not what a device-config interval means. A client waking on a fixed 15 min interval against a page whose only scheduled change is 07:00 daily burns 95 wakes to catch one change; a client on a 1 h interval against a 5 min rotation step misses most of them.

## What already exists

`app/device_upcoming.py` projects the next content change per device (`project_upcoming`), with a cause (`daily`, `interval`, `cycle`, `home_return`, `dashboard_refresh`, `widget_refresh`) and a certainty (`scheduled`, `conditional`, `estimated`). It is wired into the Companion API and the scheduler, but not into the device REST path.

## Proposed

- `_next_poll_s` consults `project_upcoming` for the device's next event and returns the delta to it, clamped to the device's configured bounds so a client is never told to wake faster than its hardware or battery budget allows.
- Fall back to the current config chain when there is no projection, or when the only candidate is `estimated`.
- Bias the returned value so the client lands *after* the render completes rather than during it, using observed render duration for that device.

## Notes

- Overlaps the per-page wake cadence item from discussion #24.
- Purely a server-side change to a value clients already honour. No protocol change, no client work, and clients that ignore `next_poll_s` are unaffected.

dmellok:
Shipped in v0.320.0: next_poll_s is pulled forward to the next projected content change, with the configured interval as the ceiling. Extended by widget-declared next_change_at in v0.322.0 (#243).

PR #314, opened 2026-09-14
feat(trmnl): dynamic /api/display refresh_rate from the shared poll decision

rpcai:
## What

The TRMNL BYOS `/api/display` path now computes `refresh_rate` dynamically, using the
same next-poll decision the v1 REST path's `next_poll_s` has used since #241. The
shared logic moves into a new `app/device_poll` module.

## Why

`app/trmnl_api.py` echoed the static configured `refresh_rate_s`. The v1 REST path
(`POST /api/v1/device/<id>/status`) has had a real decision for a while — configured
interval as the ceiling, pulled earlier for:

- the next **projected dashboard change** (`scheduler.upcoming_for_device` — schedules
  and rotation steps, quiet-window aware), and
- a **widget's own declaration** of when its data turns over (#243),

reshaped onto a **wake-alignment** grid, and **stretched to sleep through** a quiet
window the device asked to sleep through (#299). But it lived in `app/rest_api.py` and
only that path used it.

Result before this PR: a KOReader e-reader (Kobo, jailbroken Kindle) or a TRMNL panel
polling `/api/display` wakes on a blind fixed interval — every 15 min all night for a
frame that never changes, and on an arbitrary grid that misses most rotation steps.

## How

`app/device_poll.py` holds the decision, parametrised on the configured ceiling:

```python
next_poll_decision(device, *, configured_s) -> (seconds, wake_at_epoch | None)
next_poll_s(device, *, configured_s) -> int
```

- `app/rest_api.py::_next_poll_decision` is now a 3-line delegate passing
  `configured_s=_configured_poll_s(device)` (its `sleep_interval_s` resolution). Its
  behaviour is unchanged — the existing `next_poll_s` test suite passes as-is.
- `app/trmnl_api.py::display()` calls `device_poll.next_poll_s(device,
  configured_s=_refresh_rate_for(device))`, wrapped so any fault degrades to the
  static configured value. `_refresh_rate_for` stays as the ceiling / fallback.
- `display()` now logs `trmnl: served /api/display to device=%s filename=%s
  refresh_rate=%s` on every successful poll (previously only rejected/unknown-token
  polls were logged), so a poll can actually be correlated with the render it picked
  up and the cadence it was told to come back on.

The TRMNL envelope carries only `refresh_rate` (relative seconds), so the
absolute-wake-epoch half of the decision isn't surfaced there; a multi-day quiet-hours
sleep drifts a few seconds, well within the `_QUIET_SLEEP_MARGIN_S` guard, and TRMNL
firmware caps at 7 days anyway (the module caps at 6).

### Picked up a same-week upstream fix while rebasing

Between this branch's original commit and this rebase, `main` independently fixed a
race in the *REST* path's version of this same margin logic (#306,
`fix(rest): clear the scheduler tick when waking a device for a lineup`): the flat
20s content-poll margin wasn't enough to clear the scheduler's own 30s tick plus a
queued render, so a device woken right at a projected target could poll before the
frame existed and collect the previous one, then sleep its whole configured interval
on the stale frame. Their fix widened the scheduler-projected margin to
`tick(30) + 2×content_margin(20) = 70s` and tolerates a target that's overdue by less
than a tick as imminent rather than stale.

`app/device_poll.py` predates that fix, so it still had the old flat 20s margin. This
PR ports the same fix into `device_poll.py` so the BYOS path gets the identical
race-safe margin the REST path now has, rather than shipping a regression relative to
what `main` already fixed elsewhere. See `_SCHEDULER_TICK_S` / `_PROJECTED_POLL_MARGIN_S`
in `app/device_poll.py`.

## Not in scope

- **#250** (`refresh_floor_s` no longer enforced). The poll decision deliberately does
  *not* clamp to `refresh_floor_s` — that field is the glass's repaint rate, not the
  device's ask rate (a poll is a conditional GET that 304s on an unchanged frame). See
  the `_configured_poll_s` docstring. #250 belongs on the render/push path.

## Testing

- `tests/test_device_routes.py` — 4 new: projected pull-forward, configured ceiling,
  sleep-through quiet hours (app-level, no device override), projection-error
  fallback; the pull-forward assertion's expected margin was updated from the old
  flat 20s to the ported 70s tick-aware margin.
- `tests/test_rest_api.py` `next_poll_s` suite (36 tests) passes unchanged — confirms
  the refactor preserves REST behaviour, including the ported margin fix.
- Existing static-`refresh_rate` TRMNL tests
  (`test_trmnl_display_default_rate_matches_the_manifest`,
  `test_trmnl_display_serves_the_saved_rate`) pass unchanged.
- `ruff check`, `ruff format --check`, `mypy app` clean.
- Full `pytest` suite (4,870 tests, via the stacked `feat/trmnl-png-gray16` base) run
  in 8 memory-bounded batches on the dev host — all green, 0 failures. This branch's
  own targeted pair (`tests/test_device_routes.py` + `tests/test_rest_api.py`, 154
  tests) also run standalone after the margin port and the new log line — clean.
- Ran against a live instance on a Kobo Clara HD running KOReader (the `trmnl-kobo`
  koplugin): `/api/display` returns a dynamically computed `refresh_rate` that pulls
  forward to the device's own interval schedule and lands ~70s after the actual
  render, matching the ported margin exactly (confirmed against real fire timestamps,
  not just the projection math). Quiet-hours sleep-through was verified via the unit
  test above rather than live (outside the window at test time).
- **Status as of this PR**: currently under multi-day live evaluation on that Kobo.
  Initial performance looks good. Quiet-hours has *not* yet been manually confirmed
  live (only via the unit test) — will update this PR once that's checked.

This is the BYOS half of #241 — no protocol change, no client work, and a client that
ignores `refresh_rate` is unaffected.

https://claude.ai/code/session_01AAx1nPPjEnCYkuSQPwvmnx


dmellok:
Merged as 0.414.0, thank you. The lift into `app/device_poll` is a faithful move, and the REST suite passing untouched is exactly the proof I wanted. Only change on the way in: the comment above the new `/api/display` log line still said it was a temporary diagnostic, so I reworded it. The line itself stays; it is the missing half of the existing rejected-poll logging.

Two things I would like from your Kobo run when you get to them. First, whether a sleep-through quiet window actually holds the client for the whole window, since that is the one branch only the unit test covers. Second, what the smallest `refresh_rate` the KOReader plugin honours turns out to be, in case the 5 second content floor needs raising for BYOS clients. Either way, the fallback is the old static rate, so nothing is at risk while you check.

#250 stays open on the render path, as you noted.

rpcai:
I will check out the quiet hours when it triggers this evening. 

I have pivoted away from KOReader, instead now using `trmnl-kobo` ([https://github.com/usetrmnl/trmnl-kobo](url)) - KOReader was unable to put my Kobo ClaraHD into deep sleep (suspend) so battery performance was poor.

Will investigate short refreshes, but given the tmrnl-kobo approach is to wake-> connect wifi -> refresh screen -> disconnect wifi -> sleep. I think it'd be a moot point to target aynthing less than a 60 second interval for this class of devices.

PR #315, opened 2026-09-14
minor changes to German translations

bablokb:
I went through all of the `de.json` files and checked the translations.

For some of the files, I don't have the full context to really decide what is correct. Capitalization depends on various things, e.g. within lists, you tend to capitalize the first word. The same is true for column headings. Besides capitalization and the things I changed everything is fine from my perspective.

All in all, already the automated translation was probably better than what most native speakers would deliver. German spelling is a complicated thing. We even have nouns that have a different gender depending on the region, and I am not even talking about Switzerland or Austria.


dmellok:
Thank you. I appreciate you taking a look through this. I'll merge this shortly

Issue #119, opened 2026-07-17
[Bug]: Resending from history does not reset the render state (i.e. returns a 304)


bablokb:
### Tesserae version

v0.139.0

### How is Tesserae running?

Other (describe in steps)

### Operating system

Debian Trixie on MicroCloud

### Panel hardware

N/A

### Client repo + version

N/A

### Steps to reproduce

1. Send a dashboard to a device
2. Send a second dashboard to the device
3. Send the first dashboard again from the history using the "resend this push" button

The first two sends will lead to an update on the client side (HTTP 200, new link)
The third send will fail, sending a 304 

### Server logs

```shell

```

### Before submitting

- [x] I searched [existing issues](https://github.com/dmellok/tesserae/issues?q=is%3Aissue) and didn't find a duplicate.
- [ ] If I could, I tried reproducing on the latest minor release.

dmellok:
Thanks for the detailed steps.

Reproducing your exact three steps here (send dashboard A, send dashboard B, resend A from history), the resend returns 200 and the client re-fetches: A and B render to different frames, so the ETag differs. The `republish` path is unchanged since v0.139.0, so the same holds on your version.

What I did find and fix is the neighbouring case that matches your title, resending a frame whose bytes are identical to what the panel already shows. There the content-addressed ETag matches, so a REST/HTTP-polled client got a 304 and skipped the re-paint even though the resend was an explicit action (MQTT clients re-painted via force-publish; REST did not). That inconsistency is fixed in v0.149.5: a resend flags the frame so `/frame` serves one 200 to force the re-fetch, then reverts to 304 on the next unchanged poll, so deep-sleep clients still save the fetch.

If your two dashboards are genuinely different and you still see a 304 on the resend, could you share:

- the `/frame` request and response for the third send (status, `ETag`, `If-None-Match`), and
- whether the two dashboards produce visibly different frames on the panel

A server log snippet around the resend would help too. I'll leave this open until you can confirm on the latest build.

bablokb:
I still have the problem. My annotated log is below:

- first iteration is at POR, a new push (Gallery) is available, it is downloaded and displayed (BMP-name starts with 43...)
- second iteration: nothing changed, the server correctly returns HTTP304
- third iteration: new push (Calendar), the server correctly returns HTTP200, it is downloaded and displayed (BMP-name starts with 47...)
- fourth iteration: nothing changed, the server correctly returns HTTP304
[...]

PR #318, opened 2026-09-18
fix(rest): enforce a panel's refresh_floor_s on the frame-delivery path (#250)

### LINKED ISSUE #250: refresh_floor_s is no longer enforced anywhere

vaibhav8a:
Closes #250.

## What this changes

`refresh_floor_s` is enforced again — on the path that decides to hand a device a new frame, which is where the issue says it belongs.

41 hardware profiles declare a floor. Since v0.332.0 the server applied it nowhere: its only enforcement point had been the clamp on **poll** cadence in `_configured_poll_s`, and removing that clamp was right (it made an always-on E1003 wait a minute for a manual Send) but it left the repaint side unguarded.

## The rule

A frame that would land inside the floor is **held, not dropped**:

- the device is told to keep the frame it reports holding (`304`, its own `ETag`, `Content-Location` naming *that* frame rather than the one it is not getting yet);
- `next_poll_s` is pulled in to the moment the floor expires, so a device that deep-sleeps for an hour does not sit out the rest of that hour over a sixty-second floor;
- because the hold is a decision and not a queue, a render that arrives during the hold is the one that paints — the panel always lands on the latest content;
- the hold is logged with the remaining wait, because from the operator's side a deferred paint and a missed Send look identical.

Firmware needs no new behaviour: it is the `304` it already handles.

## What is deliberately not held

An unprovable floor is not a reason to withhold a frame, so these are served exactly as before:

| case | why |
|---|---|
| no declared `refresh_floor_s` | most device kinds have none |
| no recorded handover | nothing to time from; `last_served_at` only exists once a frame has been served |
| no `If-None-Match` | no frame to tell the client to keep, and a panel that booted without a URL needs one more than the glass needs the floor |
| an unchanged poll | takes the ordinary conditional-GET path, which is what keeps Companion's pending badge honest |
| a clock that went backwards | an NTP step or a container restart puts the handover in the future |

The floor check sits **ahead of** `consume_force_refetch`, so a resend (#119) caught by the floor is still a resend when the floor lifts rather than being silently swallowed.

## Shape

- `app/refresh_floor.py` — new, ~90 lines: reads the manifest field, and turns the delivery-side `last_served_at` into "seconds left", rounded up.
- `app/rest_api.py` — the hold itself, plus `_keep_current_frame()` factored out of the existing button-hold branch (#274), which was already constructing exactly this response inline.
- `app/device_poll.py` — the hold as a next-poll candidate, so both this path and the TRMNL one stay in lockstep on the cadence side.
- `tests/test_rest_refresh_floor.py` — 13 tests: the module's arithmetic, each case in the table above, and the two poll outcomes.

## Scope

The **v1 REST path only**. `last_served_at` is the closest the server has to "when the glass last moved", and it is written on that path alone — `/api/display` hands out a URL unconditionally and records no handover, so a floor there would have nothing to measure from. Recording it is a separate change; happy to follow up with one if you want TRMNL panels covered too.

## Checks

`pytest` (2109 passed; `test_mdns.py` fails locally on a missing `zeroconf`, unrelated), `ruff check`, `ruff format --check`, and `mypy app` all clean — mypy reports the same two pre-existing errors on `main` and no new ones.


vaibhav8a:
Rebased onto `main` (0.424.0 landed meanwhile) and moved this branch to **0.425.0**; #319 is on 0.426.0 so the two do not both claim a number. No change to the code itself.

dmellok:
Thanks for the careful write-up, and for keeping the change to the delivery path rather than the poll cadence. I read it against the firmware as well as the server, and I don't think we can land it in this shape. The problem is where the hold sits, not how it's built.

**1. The hold catches user-initiated repaints.**

The floor check runs after `_dispatch_query_button` / `handle_touch` have already advanced rotation state and pushed the new page, and it only tests `If-None-Match`. Firmware keeps its ETag on touch wakes (it only clears it for a physical button), so on an E1003 or Sticky a tap on a page region 20 s after the last handover does this: the server moves to page N+1, `/frame` answers 304 naming page N, the glass stays on N for up to 40 s, and any further tap in that window carries `touch_digest` = N and gets dropped as stale. The firmware comment at `main.c:363` is explicit that touch repaints count toward the floor but are never deferred by it.

Two related cases fall out of the same ordering:

- `fetch_latest` (`force_download`) is skipped past the ETag-match 304 but then caught by the hold. Unlike `force_refetch` it lives only on the transient `ButtonHandleResult`, so the press is lost, not deferred.
- A History Resend is content-addressed, so `has_pending_render` (digest != last served) is False while `/frame` still holds it. `/status` then hands back the configured interval and a deep-sleep device paints the resend an hour later, not at floor expiry. The test for this seeds `served_digest="old"` != `"same"`, so it never hits the real resend shape.

**2. The firmware already enforces this, with different numbers.**

The always-on loop holds fetched frames against `AWAKE_MIN_REPAINT_S` since last paint (`main.c:453-476`; 30 s default, 4 s on M5Paper, 60 s on OpenPaper L), and `app_config.h:247` says the guarantee lives on the device because the server never knew the panel's limit. The profile values were authored by protocol, not by glass: every ESP32/TRMNL/Xteink kind says 60, every Inky says 30, and the Sticky declares 60 while its own notes say a full paint is 1.2 s. So the server would add a serial 60 s hold from handover on top of the firmware's 30 s from paint, and an always-on M5Paper would go from repainting every few seconds to once a minute. On an always-on E1003 with a clock widget, each minute render refreshes `last_served_at`, so a manual Send is almost always inside the floor and waits for the next minute boundary. That's the v0.332.0 symptom back on the other side of the fence.

**3. Smaller things, in case you take another pass:**

- `schema/hardware.schema.json:77` and `docs/dev/adding-hardware.md:56` still define the field as a poll-cadence lower bound, so the tree would carry two meanings.
- The CHANGELOG entry says 41 profiles are enforced again, but `record_frame_served` is only called from the v1 `/frame` route, so TRMNL, `/frame.bmp` and MQTT devices get nothing. The PR body is honest about this; the changelog isn't.
- `served_at` is stamped on the JSON handover, before the download. A device whose download fails keeps its old ETag, and its retries inside the floor get 304 with no pull-in.
- The `logger.info` on the hold fires per poll; a lingering touch device polls at 1-2 s, so that's dozens of lines per hold. The firmware's equivalent log is rate-limited to once a second.
- `hold_remaining_s` already rounds up and `_floor_hold_poll_s` adds 1 again for the same stated reason.
- `test_a_newer_render_during_the_hold_is_the_one_that_paints` reseeds with `served_ago_s=61`, which also overwrites `last_served_at`, so it's testing "after expiry, newest wins" rather than a render arriving mid-hold.
- main is already at 0.424.0, so the bump is a no-op after merge.

**Where I'd take it instead**

I think the right fix for #250 is to pick one enforcement point rather than add a second. Two options, both smaller than this PR:

- Treat `refresh_floor_s` as advisory on the server (surface it in Settings and Companion, never gate on it), and let the firmware floor stand.
- Send `refresh_floor_s` in the `/status` config block and have firmware adopt `max(board, server)` locally. That keeps the hold-never-drop logic you've described in the one place that knows when the glass actually moved.

If you'd rather keep a server-side gate, it needs an intent, not just a clock: hold only scheduled / widget-change / reconcile renders and pass anything user-initiated (button, touch, resend, fetch_latest). That's a property of the render, so it would belong on the PushManager entry at stamp time rather than as an early return after dispatch has committed. Happy to talk through either direction before you spend more time on it. You may well know things about the poll path I've missed, so push back if any of this doesn't match what you've seen.

vaibhav8a:
You're right, and the design objection is the one that settles it: a second enforcement point is worse than the missing one, and I was treating the profile numbers as a property of the glass when they aren't.

I checked the parts I could check from this side, and they hold:

- **`refresh_floor_s` is authored by protocol.** `reterminal_sticky.json` declares 60 while its own `notes_md` says a full 4-gray paint is 1.2 s and a partial is under a second. Gating delivery on that number would have been slower than the panel by a factor of fifty, on a board whose file says so in the next field.
- **The resend case is real and my test never reached it.** `force_refetch` rides the existing entry, so `digest == last_served_digest`, so `has_pending_render()` is False and `_floor_hold_poll_s` returns None — held at `/frame`, no pull-in at `/status`. My test seeded `served_digest="old"` against `digest="same"`, which is a different shape entirely and passed for the wrong reason.
- **`fetch_latest` is lost rather than deferred**, since `force_download` only exists on the `ButtonHandleResult` for that request. I convinced myself the uniform rule was the honest one and didn't follow what "uniform" cost on a transient flag.
- **The tree already disagrees with itself.** `schema/hardware.schema.json:77` calls it a lower bound on poll cadence "surfaced in the Settings UI", and `docs/dev/adding-hardware.md:56` says it is "enforced on the always-on path" — the enforcement v0.332.0 removed. Nothing in `app/` reads the field at all, so the schema's own description of it is false today.

No argument on the rest either: the per-poll `logger.info`, the doubled round-up, the mid-hold test that reseeds `last_served_at` and so tests something else, and the CHANGELOG claiming 41 profiles when `record_frame_served` only exists on the v1 `/frame` route.

**Taking your first option.** Closing this in favour of the advisory shape: surface the floor where a person can see it, correct the schema and docs to describe what the field actually is, and leave the guarantee with the firmware, which knows when the glass moved and has the right numbers. That also answers the sentence in #250 that this was really about — the server should not silently carry a field it ignores — without adding a gate that fights `AWAKE_MIN_REPAINT_S`.

The `/status` option is the better end state, but the useful half of it is firmware-side and yours to sequence; happy to send the server half whenever you want it.

Thanks for reading it against the firmware. That's the context I didn't have, and it's the part that made this the wrong change.

PR #321, opened 2026-09-19
feat(hardware): state a panel's refresh floor instead of ignoring it (#250)

### LINKED ISSUE #250: refresh_floor_s is no longer enforced anywhere

vaibhav8a:
Replaces #318, taking the first of the two directions you offered there: **advisory on the server, enforcement left to the firmware.**

Closes #250 — the sentence that issue really turns on is "the server should not silently carry a field it ignores", and this stops it doing that without adding a gate that fights `AWAKE_MIN_REPAINT_S`.

## The state of the field before this

41 profiles declare `refresh_floor_s`. Nothing in `app/` read it, and the tree carried three different accounts of what it was:

| where | what it claimed |
|---|---|
| `schema/hardware.schema.json:77` | a lower bound on poll cadence, "surfaced in the Settings UI" — a surface that did not exist |
| `docs/dev/adding-hardware.md:56` | "enforced on the always-on path" — the clamp v0.332.0 removed |
| `_configured_poll_s` docstring | "belongs on the path that decides to send a new frame" — which is what sent me to write #318 |

## What this does

**States the number, gates nothing.**

- **Device card** — the floor is appended to the help on `sleep_interval_s` and `awake_poll_s`: *"This panel declares a 60s refresh floor: its firmware will not repaint the glass faster than that, however often the device asks."* Presentation only, alongside the always-on hiding in `_visible_config_fields`; the save handlers still walk the full schema and store whatever was typed. A test pins that a device set to a 5 s interval is still delivered 5.
- **Companion API** — `panel.refresh_floor_s`, nullable, so a client can explain a panel that has not changed yet. The contract entry says in as many words that it is not a promise about when the next frame lands.
- **The three descriptions** are corrected to what the field is.

## Why not enforce it here

Both of your reasons, and I checked the second one:

- The server sees a **handover**, never a paint. `served_at` is stamped when the JSON goes out, before the download; a floor timed from it is measured from the wrong instant.
- The values are **authored by protocol, not by glass**. `hardware/seeed/reterminal_sticky.json` declares `refresh_floor_s: 60` and its own `notes_md`, two fields later, puts a full 4-gray paint at 1.2 s and a partial at under a second. Gating on that would have been fifty times slower than the panel, per the panel's own file.

Firmware holds its repaints against a floor measured from the last paint. That is the one place that knows, and it keeps the guarantee.

## Tests

`tests/test_refresh_floor_surfaced.py`, 6 cases: the note reaches both cadence fields and names whose guarantee it is, it keeps the help the field already had, a kind that declares no floor gets nothing added, the setting is still stored and delivered unclamped, the Companion value is the profile's (and `None` for a floorless kind), and the schema and guide no longer describe a poll-cadence bound.

## Checks

`ruff check`, `ruff format --check`, `mypy app` clean (same two pre-existing errors as `main`, none new). 624 passed across the settings / device / REST / companion suites, including the Companion contract validation — which caught the new field until the OpenAPI `Panel` schema was updated, so that is in here too.

The `/status` route is the better end state and I'd still like to send the server half whenever the firmware side is on your list.


dmellok:
Thanks for taking the advisory direction, and for keeping the code change small. Four things before I can land it, then some nits.

1. The note promises something the firmware does not do. The firmware holds `AWAKE_MIN_REPAINT_S`, a compile-time constant it never reads from the profile (30 s default, 4 s on M5Paper, 60 s on OpenPaper L), and only in the always-on loop. Deep-sleep wakes paint every time the ETag changes, and the Pi, Pico, PicPak and TRMNL clients have no floor code at all. So "its firmware will not repaint faster than 60s, however often the device asks" is wrong on an E1003 in both modes, and the same claim sits in the contract description, schema, docs row and the `_configured_poll_s` docstring. Could you word it as what the profile declares ("the profile lists a 60 s repaint floor for this glass") and not name the firmware as the guarantor anywhere? The test that pins "firmware" into the help text should change with it.

2. The vendored contract should stay verbatim. Please move the `Panel.refresh_floor_s` addition into `tests/companion/contract_errata.py` as a `SchemaErratum`, like the three existing entries. A re-vendor would otherwise drop it and fail the devices listing with no pointer back.

3. The two `trmnl_client` profiles that declare a floor use `refresh_rate_s`, so the note never shows for them. Adding it to the cadence tuple, plus a test on one of those kinds, would cover it.

4. `test_the_note_is_presentation_only` registers `esp32_client`, which has no floor, so a reintroduced clamp would still pass. `seeed_reterminal_e1003` would make it real.

Nits: one `Device.refresh_floor_s` property instead of the two copies of the accessor; the Companion fixtures could carry the field in both forms; `encoding="utf-8"` on the two `read_text` calls; and I keep prose free of em-dashes, there are six new ones. The help lands in the "?" popover rather than inline, so "shown beside" is a little generous in the changelog.

The pytest failure on CI is the deck-card "poll ≈" assertion, which passes locally on your branch and is unrelated. Needs a rebase for CHANGELOG/pyproject/uv.lock; 0.426.0 is still free.

vaibhav8a:
Thanks for the thorough read. All four are done, plus the nits. Rebased onto 0.427.3 and bumped to **0.428.0**, since 0.426.0 was taken while this sat.

**1. No guarantor named.** The note now reads *"The profile lists a 60s repaint floor for this glass."* The same framing is in the contract description, the schema, the guide's row and the `_configured_poll_s` docstring: a declaration, not a guarantee, with no mention of firmware. The test that pinned "firmware" into the help text now asserts it's absent, and the schema/guide test does the same.

**2. Contract verbatim.** `Panel.refresh_floor_s` is a `SchemaErratum` in `contract_errata.py`, and `app-v1.openapi.yaml` is byte-identical to `main` again. `test_each_schema_erratum_is_still_needed` covers it, so the first re-vendor that brings it in will name the entry to delete.

**3. TRMNL kinds.** `refresh_rate_s` is in the cadence tuple. I asked the registry which floored kinds carry that field and got exactly the two you meant, `trmnl_x` and `seeed_xiao_75`. There's a test on `trmnl_x`, and it fails if `refresh_rate_s` comes back out of the tuple.

**4. A real presentation-only test.** It registers `seeed_reterminal_e1003` now. To check it catches the thing it's for, I clamped `_current_config` to the floor, and it failed with `assert 60 == 5`. The `esp32_client` version would have passed.

**Nits**
- One `Device.refresh_floor_s` property. Both the settings note and the Companion view read it, and the two copies are gone.
- Fixtures: the shared ones sit inside the vendored `contract/` directory, so editing them would undo point 2. Instead, a local test validates a `Panel` against the erratum-patched schema with the field as `60` and as `null`, and checks that `0` is refused. If you'd rather have it in a fixture, point me at where you keep local ones and I'll move it.
- `encoding="utf-8"` on both `read_text` calls.
- No em-dashes in anything this PR adds. I checked the added lines of the diff, and there are none.
- The changelog now says the note is in the "?" help.

The deck-card "poll ≈" failure you mentioned didn't come up in the 785 tests I ran across the settings, device, poll and hardware suites, so we'll see whether CI agrees.

dmellok:
Thanks, this is the shape I was hoping for. All four checked out against the branch:

1. The note, the erratum description, the schema and the guide row all source the number to the profile and name no guarantor, and the tests pin "firmware" out of the help text.
2. `app-v1.openapi.yaml` is byte-identical to main, and the `Panel.refresh_floor_s` erratum sits in `contract_errata.py` where the still-needed check will name it when a re-vendor catches up.
3. `refresh_rate_s` in the cadence tuple, with `trmnl_x` covered.
4. `seeed_reterminal_e1003` in the presentation-only test, and the clamp check you ran is the right way to prove it bites.

Keeping the local schema check in the test rather than a fixture is fine; the fixtures are vendored for the reason you gave.

main moved on to 0.427.6 while this sat, so the only conflicts were the changelog, `pyproject.toml` and `uv.lock`. I resolved those (0.428.0 stands, and the entry now sits under the existing "Changed" heading rather than a second one) and merged. The deck-card "poll ≈" failure didn't reappear here either, so that was the flake it looked like.

PR #324, opened 2026-09-20
cloud relay deck navigation fix

varanu5:
## What this changes

Cached lineup navigation now notifies delivery listeners after selecting the frame, allowing the relay publisher to upload it. Background warming stays silent, and unchanged frames remain deduplicated.

## Why

I found that PicPak’s next-page button worked through the cloud relay with lineup background refresh disabled, but stopped working when it was enabled. The same lineup worked over REST.

When the next page was already cached, `promote_deck_page()` updated the server’s live frame without notifying the relay publisher. REST could serve the selected frame immediately, but the relay mailbox retained the previous image.

I added a `PushResult` and a call to `_notify()` after releasing the push lock, following the normal push path. The production change is seven added lines in `app/push.py`.

In the failing device log, the button waiting window opened normally, but every fetch returned 304. After this change, the server uploaded the selected frame and PicPak received and painted it without a firmware update.

## Test plan

- [x] Added a regression test that fails before the fix and passes afterward. It verifies that:
  - Background warming stays silent.
  - A cache miss does not notify delivery listeners.
  - Cached-page promotion notifies listeners outside the push lock.
  - The uploaded ETag and decrypted frame bytes match the selected frame.
  - Selecting identical frame bytes does not upload them again.
- [x] Ran the following tests: **168 passed**.

```bash
pytest -q -p no:cacheprovider \
  tests/test_deck_prerender.py \
  tests/test_relay_home.py \
  tests/test_deck_nav.py \
  tests/test_deck_timer.py \
  tests/test_deck_refresh.py \
  tests/test_rest_deck.py \
  tests/test_push.py \
  tests/test_opendisplay_ha.py \
  tests/test_ha_discovery.py
```

- [x] Checked lint, formatting and whitespace for the changes.

```bash
ruff check app/push.py tests/test_deck_prerender.py
ruff format --check app/push.py tests/test_deck_prerender.py
git diff --check
```

- [x] Manually verified on PicPak hardware with background refresh set to **15 minutes** and unchanged firmware. The uploaded frame ID matched the device’s received ETag. PicPak stopped polling, painted and returned to sleep.
- [ ] Full test suite: `pytest -q` — not run.
- [ ] Repository-wide lint and formatting: `ruff check . && ruff format --check .` — not run.
- [ ] Type checking: `mypy app` — not run.

The automated regression test mocks browser capture and network uploads. I verified actual relay delivery and display refresh separately on hardware.

## Checklist

- [x] Regression test added.
- [ ] Repository-wide `ruff` and `mypy` checks completed.
- Documentation: unchanged; this fixes existing behaviour without changing setup or configuration.
- [ ] CHANGELOG entry under `## [Unreleased]` — not added.
- Version: unchanged; this PR does not prepare a tagged release.

Only `app/push.py` and `tests/test_deck_prerender.py` are modified.


dmellok:
Merged, thanks. The diagnosis is right: `promote_deck_page()` was the one live-frame swap on the deck path that never told the push listeners, and since the relay upload hangs off that listener, the mailbox kept the previous image while REST served the new one straight from the live slot.

Worth saying that this reaches further than the press. The same promote advances a deck on its timer, returns it to its home page, and backs touch navigation, the companion app and the Decks page controls, so relayed panels were missing all of those too. I've written the changelog entry that way.

Two notes from the review, neither of which held up the merge:

- The result you synthesise carries no renderer row, and Home Assistant discovery walks that list to reach each display's own entities. A cached step moved the hub-level "active dashboard" but left the navigating panel's current-dashboard, frame and last-updated entities on the page it had just left. Follow-up in 0.426.2 fills the row in from the cached render info.
- `_promote_shadow()` looks like the same gap. The patch path promotes a shadow render into the live slot when patching isn't the right delivery, again without notifying, so a patch-capable relay panel probably won't get that frame uploaded either. I haven't confirmed it on hardware. If you want it, it's the same shape as this change.

Putting `_notify()` outside the push lock is right, and the test pinning that is appreciated: the publisher calls back into the manager, so notifying under the lock would deadlock. The hardware confirmation with background refresh on is the part I couldn't have checked from here.

